PR #9, opened 2019-02-20
Support OSX

RetroFloppySupport:
A couple of tweaks to build on OSX.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/.project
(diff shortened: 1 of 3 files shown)


davidgiven:
Hello --- thanks for this. Unfortunately it breaks the build on Linux, because the local endian.h is found in preference to the system one.

I did actually last night add (as part of the Amiga stuff) a native byte reader/writer library, so I don't actually need <endian.h> at all any more; accordingly, I've taken it out. The FluxEngine client code now builds happily on OSX and I've updated the Travis environment to do so automatically. Thanks for prompting me into doing that!

PR #17, opened 2019-02-22
File sector start addresses are 16 bits long (well, maybe only 9)

david-schmidt:
Brother 120k starting sector addresses are actually two bytes long.  I'm sure there are prettier C++ish ways to express this, but there you go.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.obj
+.project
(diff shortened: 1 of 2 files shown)


davidgiven:
Yup, there are, but this is perfectly good enough. Thanks!

Issue #37, opened 2019-03-06
building the client under FreeBSD fails


tingox:
Hello. I'm trying to build the client under FreeBSD, but the process fails, like this:
'''
tingo@kg-core1$ make
meson .obj
The Meson build system
Version: 0.49.2
Source dir: /zs/tingo/personal/projects/PSoC/fluxengine
Build dir: /zs/tingo/personal/projects/PSoC/fluxengine/.obj
Build type: native build
Project name: fluxclient
Project version: undefined
Native C++ compiler: c++ (clang 6.0.1 "FreeBSD clang version 6.0.1 (tags/RELEASE_601/final 335540) (based on LLVM 6.0.1)")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/local/bin/pkg-config (1.6.0)
Dependency libusb-1.0 found: YES 1.0.13
Dependency sqlite3 found: YES 3.27.1
Dependency zlib found: YES 1.2.10
Build targets in project: 47
Found ninja-1.8.2 at /usr/local/bin/ninja
ninja: Entering directory `.obj'
[17/129] Compiling C++ object 'fluxreaderlib@sha/lib_fluxreader_sqlitefluxreader.cc.o'.
FAILED: fluxreaderlib@sha/lib_fluxreader_sqlitefluxreader.cc.o 
c++ -Ifluxreaderlib@sha -I. -I.. -I../lib -I../dep/fmt -I../lib/stream -Xclang -fcolor-diagnostics -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -g --std=c++14 -fPIC  -MD -MQ 'fluxreaderlib@sha/lib_fluxreader_sqlitefluxreader.cc.o' -MF 'fluxreaderlib@sha/lib_fluxreader_sqlitefluxreader.cc.o.d' -o 'fluxreaderlib@sha/lib_fluxreader_sqlitefluxreader.cc.o' -c ../lib/fluxreader/sqlitefluxreader.cc
In file included from ../lib/fluxreader/sqlitefluxreader.cc:3:
../lib/sql.h:4:10: fatal error: 'sqlite3.h' file not found
#include <sqlite3.h>
         ^~~~~~~~~~~
1 error generated.
[18/129] Compiling C++ object 'fluxreaderlib@sha/lib_fluxreader_hardwarefluxreader.cc.o'.
[...]

PR #44, opened 2019-03-11
fix meson.build so that fluxengine compiles under FreeBSD

tingox:
Necessary changes to meson.build that makes fluxengine build under FreeBSD.


tingox:
This fixes #37 .

davidgiven:
Awesome, thanks. (I did get my FreeBSD box set up... but something's wrong with the update and package servers so I was unable to install a development system.)

PR #97, opened 2019-08-20
CFLAGS exported all through

hpingel:
Fixing make on Ubuntu 18.04 amd64 as it could not find libusb.h.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ export LIBS = $(shell pkg-config --libs $(PACKAGES))
 export EXTENSION =
 endif
 
-CFLAGS += -Ilib -Idep/fmt -Iarch
+export CFLAGS += -Ilib -Idep/fmt -Iarch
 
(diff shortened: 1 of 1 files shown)


davidgiven:
That shouldn't be necessary; CFLAGS is defined as being exported a few lines above. (In both branches of the if statement.) I've just verified that appending something to a variable doesn't unexport it, so unless the behaviour of make has changed you may have modified something else...

hpingel:
Okay, here is what happened: (this pull request is obsolete)

On Ubuntu 18.04 I did "make all". Then it complained that libsqlite3 (libsqlite3-dev) was missing so I installed it via apt. Then I did "make all" again without "make clean" and then it complained about the missing "libusb.h". But libusb-1.0.0-dev was installed all the time.

Then after my change to the Makefile  I must have done make clean and make all again and then it worked. But it would have worked without the change of the Makefile.

Issue #138, opened 2020-02-20
Missing sector doesn't force a retry.


Gandalf-ND:
I've been using FluxEngine to image a number of Norsk Data floppies, both 5.25" DS-HD and 8" SS-DD. (Yeah, I'm a friend of Tingo.) The most common problem I've run into is "Missing sectors". Rerunning the floppy often results in the sector found and in most cases also other sectors missing. In the floppies I copy I have had this issue in about every fourth floppy... and I've done almost 500 floppies so far.

Digging into the code I've traced it down to a failure of detecting the sector header. If the sector header is there but the data record is not detected it leads to a "Sector is there but empty" and a repeat. If the sector header is missing then the program just skips until it detects the next sector header and reads the next sector. If it is detected properly we don't get an error. Now there is a hole in the sector table and FluxEngine fails to see a problem.

I don't know how to best detect a missing sector. Would it be possible to have an option with --ExpectedSectors=8 (for example in my case) and do a test after a track read that we really do have the expected number of tracks, and if we don't have the expected number of sectors then we read the track again up to the number of repetitions needed.

I was looking through the code but it's been a number of years since I wrote any c++ so I don't feel confident enough to suggest a patch, at least not yet. I'm thinking of adding a quick and dirty hack to /lib/reader.cc in readDiskCommand() just to test my theory.

I will put the images and flux data up on a web server soon.

Thanks for a great project! :-)

davidgiven:
As you say, telling the difference between a missing sector and one which is supposed not to be there is an issue. I don't think there's any solution other than to tell the reader what sectors to expect. It'd be easiest to add this to `reader.cc` but I think it actually makes more sense in the `arch` format-specific code, as that's the bit of code which actually knows.

It's a bit nasty; it would require adding a method to AbstractDecoder which would return a set of sector IDs, which `readDiskCommand()` would query... but it'd be fairly easy to implement.

BTW, how are you getting on with the ND disks? I've recently done a whole bunch of changes and bugfixes which should make reading disks massively more reliable (i.e. fixing the horrible buffer overrun in the firmware). Unfortunately both my 5.25" drives have died and I'm unable to test it on the ND disks which tingox sent me.

Also, it should be possible to write ND disks now, with care.

Gandalf-ND:
[...]

Issue #82, opened 2019-08-02
Autodetect drives on the bus


davidgiven:
Remember how old PCs would grind their disk drives on startup, as they do disk seeks? This tells me that drives support seeking, and track 0 detection, even with no disk in the drive. This should allow us to detect whether drives are attached to the bus. This is particularly valuable as it would allow a single drive to be treated as drive 0, regardless of whether the FluxEngine is attached in piggyback mode or PC drive cable mode, and avoid the nasty numbering issue we currently have.

PR #139, opened 2020-02-21
Add support for required sectors, drive autodetection and fix homing on 8" drives.

davidgiven:
- add support for giving the reader a set of required sectors, so if one is missing then we can tell and the track can be reread
- detect drives at poweron, so a single drive will always be drive 0 regardless which connector it uses
- don't seek to track -1 when homing because it upsets 8" drives


davidgiven:
Fixes: #138 
Fixes: #82

PR #162, opened 2020-05-10
flipped SVG track order to put 0 at the outside

flimshaw:
I am not sure if this is a universal bug or anything, but it seemed to me that the render order of tracks on the svg output was backwards for my Brother WP disks, so I figured I'd send a PR.


davidgiven:
Yes, indeed it is. This explains why I was never able to get a good visualisation on C64 disks.

hharte (approved):

PR #164, opened 2020-05-20
Simplify Sampler.

jboone:
I've simplified the sampler by:

* Removing the state machine, which was unnecessary to drive the FIFO enable signal.
* Mading the `req` signal registered.
* Removing the pulse stretchers on the RDATA and INDEX signals. I think they're unnecessary for pulses that should always be >83ns (1 / 12 MHz).
* Simplifying capture of rising edges of RDATA and INDEX to always update them at each sample clock, not setting them in one place and clearing them in another.
* Using Verilog net-naming practices consistent with other Verilog projects I've worked with, including suffixes "_q<n>" for denoting delayed, registered versions of signals.

I'm not doing this just to get my stink on the code. ;-) This had an enormous impact on readability of 800K Mac floppies I've been trying to read.

Before ("Clean and Build", "Release" using PSoC Creator 4.3, git commit 3c54a663b87629b57907c8b691cb268b6f80a1d5):

```
H.SS Tracks --->
0. 0 BX???B?XX?XX?XX?XB??X?B??BXBXBBB................BBBBBBBBBBBBBBBB..........B....B
0. 1 XBXX???B?XXXXXX?X??B?X?BBBBBBBBB................BBBBBXBBBBBBBBBB................
0. 2 XBX?X?B?B???X?X?XBXXB?X?BBBBBBBB................BBBBBBBXBBBBBBBB..........B..B..
0. 3 XBX??X?XX?X?XX??BBB?BBBBB??BBBBB................BBBBBBBBBBBBBBBB..........B...B.
0. 4 ?X?X?XBX?XBX??X?BBBBB?BBBBXBBBBB................BBBBBXBBBBBBBBBB.........B...B..
0. 5 X?BXX?XXX?BX??XXXX??BXBXBBBBBBXB................BBBBBXBBBBBBBBXB.........B....B.
0. 6 ?XX???XXX??XX?XXBXXBBBBBBBXBBXXB................BBBBBBBBBBBBBBBB.........B....BB
0. 7 BXXB??BXX?XXXXX?BXX?BBBBBXBBXBBX................BBBBBBBBBBBBBBBB.........B..B..B
0. 8 XXX???XXXXXXXXX?XXBXX?XBXBBXBBXX................BBBBXBBBBBBBBBBBXXXXXXXXXXXXXXXX
0. 9 ?XXX??XXXBXXX?XXB?XB?BBBBBBXBBXB................XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
0.10 ?B?XBXX?XXX?B?X?BX?XBBBXXBBBBBBBXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
0.11 X?XB?XXXB??XXX?XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
Good sectors: 363/960 (37%)
Missing sectors: 310/960 (32%)
Bad sectors: 287/960 (29%)
writing 80 tracks, 1 heads, 12 sectors, 524 bytes per sector, 491 kB total
```

After:

```
H.SS Tracks --->
0. 0 ..........................................................................B....B
0. 1 .............................................................................B..
0. 2 ..........................................................................B..BBB
0. 3 ..........................................................................B...B.
0. 4 .........................................................................B...B..
0. 5 .........................................................................B....B.
0. 6 .........................................................................B....BB
0. 7 .........................................................................B..B..B
0. 8 ................................................................XXXXXXXXXXXXXXXX
0. 9 ................................................XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
0.10 ................................XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
0.11 ................XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
Good sectors: 781/960 (81%)
Missing sectors: 160/960 (16%)
Bad sectors: 19/960 (1%)
writing 80 tracks, 1 heads, 12 sectors, 524 bytes per sector, 491 kB total
```
I've repeated the test a couple of times, and it's consistent and night-and-day. I also went back and read an IBM DD and HD floppy to increase my confidence I wasn't regressing other formats.

If I wasn't lazy, I'd simulate the old and new code with iVerilog and GTKWave and try to understand why the improvement. But that would involve writing a testbench...

Anyway, give it a shot and see if it doesn't help!

And finally, thanks for this excellent and clever project! I've been using the hell out of it. :-)


davidgiven:
Hello --- that's amazing! Thank you very much! I've been wondering about this for ages. I haven't had a chance to look at it yet due to being busy doing other things but should get onto it soon.

FluxEngine.cydsn/Sampler/Sampler.v
```diff
@@ -19,77 +18,60 @@ module Sampler (
...
         /* Remember positive egdes for sampleclock, index and rdata. */
         
-        if (sampleclock && !oldsampleclock)
-            sampleclocked <= 1;
-        oldsampleclock <= sampleclock;
-        
-        if (index && !oldindex)
-            indexed <= 1;
-        oldindex <= index;
+        /* Positive-going edge detection of 16 MHz square-wave sample clock vs.
```
davidgiven, line 47: The sample clock is 12MHz, actually...

davidgiven:
Sorry this took so long to get too -- looks good to me! I can confirm that it's working drastically better. I'm still not getting clean reads, but the weird banding has gone away. I don't have a Mac and my disks are probably bad. I'm still not convinced that there isn't a problem with my floppy drive and Mac disks.

I've left one comment.

I am curious to know why my sampler code didn't work, though (apart from my Verilog clearly being terrible). The sample clock is 12MHz, which means there's 5 state machine ticks between samples, which should be plenty of time to work through the state machine. Not having a state machine at all is obviously preferable.

I can see one potential edge case which is as there's no state machine there's no way to force the FIFO to relax between opcodes. So, if events happen on two consecutive state machine ticks, then the second one will be lost. This could happen if, say, an rdata edge happens on one tick and then an index edge happens on the next. I think this'll be vanishingly rare, though.

davidgiven:
BTW, the pulse converters on the data inputs were primarily there to synchronise the pulses with the sample clock. As the FDD is fundamentally an analogue device they could arrive at any point. This was recommended to me as a Verilog newbie to avoid problems.

I wish the Cypress tooling had some kind of simulation or test suite system...

FluxEngine.cydsn/Sampler/Sampler.v
jboone: Oh yeah, sorry, I knew from working from .flux files that the counter rate is 12 MHz. That comment's wrong though I don't think it has any impact on the code.

jboone:
> I am curious to know why my sampler code didn't work, though (apart from my Verilog clearly being terrible). The sample clock is 12MHz, which means there's 5 state machine ticks between samples, which should be plenty of time to work through the state machine. Not having a state machine at all is obviously preferable.

I've been wondering this too, and really need to simulate it to tell for sure. My wild-guess-hypothesis is that, when the input pulse rate is approximately the counter overflow rate, you get wrong output somehow. But I can't point to anything clearly wrong. A simulation would tell for sure.

OK, maybe I'll try that tonight. :-)

jboone:
> BTW, the pulse converters on the data inputs were primarily there to synchronise the pulses with the sample clock. As the FDD is fundamentally an analogue device they could arrive at any point. This was recommended to me as a Verilog newbie to avoid problems.

I think the danger from noise is quite low. The signals themselves and the rising edges are very slow, especially with the light pull-ups you're using, and you have input hysteresis (Schmitt trigger) enabled, if I recall correctly. And I believe the GPIO architecture on the PSoC has a two flip-flop synchronizer on every pin by default. But experiment with keeping them in! I'm not sure removing them had any impact. I mostly took them out as an effort to remove anything I didn't understand completely before I tried to make improvements.

jboone:
Suspicion confirmed!

Screenshot from 2020-05-27 20-33-29

In the top section are the test signals driving both implementations. Below that, I've got the "Sampler (Before)" and "Sampler (After)" code simulating side by side. It might be a little hard to see, but in the "before" section there's a pulse interval (333 system clocks @ 64 MHz, or 5.2 microseconds) where opcode 0x3F is emitted twice in a row. If you're working with a floppy format that produces pulses with that duration (or a multiple, I expect), you'll wind up with a lot of extra 5.2 us gaps in the data.

I've output the opcodes during the simulation to a file ([before.txt](https://github.com/davidgiven/fluxengine/files/4693175/before.txt) and [after.txt](https://github.com/davidgiven/fluxengine/files/4693178/after.txt)) so you can see what happens as the pulse interval approaches and passes through 5.2 us.

```
bd
be
be
be
be
be
be
bf
bf
bf
bf
3f <--- one of these
3f <--- shouldn't be here.
81
3f
81
3f
81
3f
81
3f
82
3f
81
3f
```
I used a little chunk of Python to recover timestamps from the opcodes and plot the results:

```python
#!/usr/bin/env python3

import sys

def opcodes_to_timestamps(file_path):
	timestamps = []

	t = 0
	with open(file_path) as f:
		for line in f:
			v = int(line.strip(), 16)
			delta = v & 0x3f
			t += delta

			if v & 0xc0:
				timestamps.append(t)
				t = 0

	return timestamps

timestamps_before = opcodes_to_timestamps('before.txt')
timestamps_after  = opcodes_to_timestamps('after.txt' )

import matplotlib.pyplot as plt

plt.plot(timestamps_before, label='Before')
plt.plot(timestamps_after,  label='After' )
plt.legend(loc='upper left')
plt.xlabel('System Clock Ticks (64 MHz)')
plt.ylabel('Sample Ticks (12 MHz)')
plt.show()

```

Screenshot from 2020-05-27 21-16-46

The new code addresses these spikes. However, I'm noticing some other things I'm not sure are OK with both versions of the code... Like a bit of a DC bias vs. the theoretically correct pulse intervals, and a scale error of 2.7%. Floppies tend to be pretty squishy as it is, but 2.7% might be enough to cause a problem. I need to see if this is a simulation or graphing error on my part, first...

jboone:
I changed my testbench to track absolute time instead of deltas, so it was less difficult to tell if I was accumulating errors over time. Sure enough, it appears there's a significant accumulation of ticks over time.

Screenshot from 2020-05-27 22-14-24

I wonder if this is a rounding issue. I retried using a system clock of 12 MHz, and it produced perfect results.

Screenshot from 2020-05-27 22-22-23

I thought about it a bit more, and I think the right solution here, to cure all ills, is to do virtually all of the sampler code in the sampling clock domain, but then synchronize into the system domain when you want to feed the FIFO. Only the `req` signal will exist in the system clock domain, and it'll be a single clock pulse wide. I'll try hacking something together...

jboone:
OK, done.

```verilog
// NOTE: Reset pulse is used in both clock domains, and must be long enough
// to be detected in both.

reg [5:0] counter;

reg index_q;
reg rdata_q;

reg index_edge;
reg rdata_edge;

reg req_toggle;

always @(posedge sampleclock)
begin
    if (reset)
    begin
        index_edge <= 0;
        rdata_edge <= 0;
        index_q <= 0;
        rdata_q <= 0;
        counter <= 0;
        req_toggle <= 0;
    end
    else
    begin
        /* Both index and rdata are active high -- positive-going edges
         * indicate the start of an index pulse and read pulse, respectively.
         */
         
        index_edge <= index && !index_q;
        index_q <= index;
        
        rdata_edge <= rdata && !rdata_q;
        rdata_q <= rdata;
        
        if (rdata_edge || index_edge || (counter == 6'h3f)) begin
            opcode <= { rdata_edge, index_edge, counter };
            req_toggle <= ~req_toggle;
            counter <= 1;
        end else begin
            counter <= counter + 1;
        end
    end
end

reg req_toggle_q;

always @(posedge clock)
begin
    if (reset) begin
        req_toggle_q <= 0;
        req <= 0;
    end else begin
        req_toggle_q <= req_toggle;
        req <= (req_toggle != req_toggle_q);
    end
end
```

I like this code better. It's a bit simpler, and more rigorous, I think. It confines the clock domain crossing to a very tiny bit of code -- just the part in the `always @(posedge clock)` block. The synchronization technique is a simple toggle. The slow domain toggles a flip-flop each time something interesting happens. The fast domain watches the output of the flip-flop and compares it with a delayed version. If there's a change, it produces a pulse. We get at least one fast domain clock cycle worth of setup time going into the FIFO.

I've only tested this in simulation, not on actual hardware. But I anticipate it will work fine.

Unfortunately, this didn't help with the scale error. I'm back to thinking this is a rounding problem. But I've run out of time to think about it tonight.

jboone:
By the way, I found these resources to be excellent for learning more about various clock domain crossing problems and techniques for solving them:

[Clock Domain Crossing Design – 3 Part Series](https://www.verilogpro.com/clock-domain-crossing-part-1/)

[Crossing the abyss: asynchronous signals in a synchronous world](https://www.edn.com/crossing-the-abyss-asynchronous-signals-in-a-synchronous-world/) -- available as a PDF I have squirreled away for posterity.

davidgiven:
Thank you for the analysis --- graphs make everything better. What simulation tool did you use? Because if I take one thing home from all this, it's that I need to learn how to simulate.

The new code works absolutely fine, It isn't obviously better with regard to reading dubious disks, but I like that it's small, and simple, and contains a number of interesting features that I wouldn't have thought of...

I'm happy to submit this. Do you want to update the pull request or shall I just patch it in? I tried it with the pulse converters both present and missing, and they make no difference either, so I'll go for the simpler option and remove them, 

Re the scale error: 2.7% is more than I'd expect, yes. The old code had similar problems, though. Eventually I bodged around it. I am suspicious of your `counter <= 1`; this means that the last tick of one intervals is also the first tick of the next interval. For a DD disk with intervals of about 55 ticks, then one tick is a bit more than 2%, which is a suspiciously similar number...

jboone:
> Re the scale error: 2.7% is more than I'd expect, yes. The old code had similar problems, though. Eventually I bodged around it. I am suspicious of your `counter <= 1`; this means that the last tick of one intervals is also the first tick of the next interval. For a DD disk with intervals of about 55 ticks, then one tick is a bit more than 2%, which is a suspiciously similar number...

Heh, I got that from your code! And initially I questioned it too. But you left a revealing comment (that I didn’t copy over to my code):

```verilog
counter <= 1; /* remember to count this tick */
```
...and in thinking about it, I determined it you were correct. My thought process was that, on the clock cycle where an opcode is recorded, that would be the zeroth clock cycle to the next tick. In other words, if you produced several consecutive opcodes with a zero delta, they’d all occur at the same time. So the next time that code is executed (the next tick), if another opcode were to be generated, the opcode should contain a delta of one — one cycle since the last opcode, right? So the counter should be set to the value that would need to be in the next code after the one we’re recording.

Keep in mind the nature of synchronous logic as described in HDLs like Verilog. All the assignments (rvalues) don’t take effect until *after* the clock edge, while all the variables (lvalues) are evaluated before the clock edge.

jboone:
> Thank you for the analysis --- graphs make everything better. What simulation tool did you use? Because if I take one thing home from all this, it's that I need to learn how to simulate.

I’ll drop my simulation in a separate branch for you to take a look at. It doesn’t meld with the way the Cypress code generation works, so I had to modify the code.

I used `iverilog`, `vvp`, and `gtkwave`. And a small pile of Python and libraries.

jboone:
I've added my simulation files and a README to a branch in my repo:

https://github.com/jboone/fluxengine/tree/sampler_verilog_sim/rtl

davidgiven:
I figured out how to merge this manually, and still keep your username on the Verilog file. It all seems to have worked. Thank you very much!

PR #191, opened 2020-10-03
Add Micropolis decoder

### LINKED ISSUE #187: 5¼" hard-sectored 100 TPI CP/M floppies for Vector 4
I have a working Vector 4 made by Vector Graphics Inc in the early 1980s. It uses a 5¼" 16 hard sector sector floppy with 256 byte sectors and 100 TPI, I believe. The user guide claims a capacity of 630 KB which would mean 77 tracks (although [I see](http://retrotechnology.com/herbs_stuff/d_micropolis.html) a reference to an 80 track floppy being used as 77). The disk drive itself has a sticker that says "100 TPI DSR," although we purchased 96 TPI floppies to use with it. Earlier versions of the Vector seemed to use the Micropolis floppy disk controller and I think this one is probably compatible, so 100 TPI makes sense. This [brochure for Micropolis](https://web.archive.org/web/20170325221014/https://amaus.net/static/S100/micropolis/brochure/micropolis%20metafloppy.pdf) would make me believe it probably uses MFM encoding.

I've dug into CP/M and am semi-close to getting disk images using the Vector 4 itself. Getting images of the majority of the data is fairly easy, but reading the system track is more difficult and I'm still working on that. With a bit of work I can also dump the CP/M disk parameter header+block that has things like logical to physical sector mapping, logical sectors per track, and reserved system tracks. I'd be able to write a custom pattern to the disk, except for the system track.

What is the best way to go about getting these disks readable with fluxengine? The ds990 encoder looks like it was reasonably straight-forward, although I'd definitely need to guess-and-check and would need to consult a hard-sectored decoder. `fluxengine rpm` shows 6000 rpm, which is weird as I'd expect it to be closer to 5100 rpm (300 * 17, for 16 sector holes + 1 index hole). But it does seem to be at least semi-working as 6000 isn't completely outrageous and `inspect --dump-flux` isn't all zeros.

I've got the original Vector 4 floppy drive and a DOS-era floppy drive that seem to be working with fluxengine. Given the 100 TPI, it would seem that I need to use the original Vector 4 floppy drive and not the DOS-era one. Does that sound right?

ejona86:
Resolves #187

### DIFF
--- a/arch/micropolis/decoder.cc
+++ b/arch/micropolis/decoder.cc
@@ -1,0 +1,56 @@
+#include "globals.h"
+#include "fluxmap.h"
+#include "decoders/fluxmapreader.h"
+#include "decoders/decoders.h"
+#include "sector.h"
+#include "micropolis.h"
+#include "bytes.h"
+#include "fmt/format.h"
+
+/* The sector has a preamble of MFM 0x00s and uses 0xFF as a sync pattern. */
+static const FluxPattern SECTOR_SYNC_PATTERN(32, 0xaaaa5555);
+
+AbstractDecoder::RecordType MicropolisDecoder::advanceToNextRecord()
+{
+	_fmr->seekToIndexMark();
+	const FluxMatcher* matcher = nullptr;
+	_sector->clock = _fmr->seekToPattern(SECTOR_SYNC_PATTERN, matcher);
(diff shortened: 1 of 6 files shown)


davidgiven:
Sorry for the delay --- this looks perfect! You even wrote documentation!

Next stop: write support...

PR #200, opened 2020-12-25
Add firmware support for hard sectors

ejona86:
Feel free to rip this apart. I would have discussed first, but things sort of fell into place and then it actually worked the first time it compiled...

Write support is untested, but that is the purpose and real goal. I hadn't added write support for Micropolis because the writes need to be aligned to the start of the track. I have not tested with retro hardware, but I have tested with the Micropolis support I added to FlashFloppy with some local changes to improve index timing (based on logic analyzer). It doesn't really impact reads much, but it appears to be working as `rpm` now produces consistent and correct results and things like `--sync-with-index true --revolutions .9` produce properly-aligned errors.

It'll be at least a month until I have access to Micropolis hardware again, for further testing and verification.

Edit: I've now tested writing for Micropolis (future PR), and the write stream is properly synchronized to the start of the track when viewed with a logic analyzer.

### DIFF
--- a/FluxEngine.cydsn/main.c
+++ b/FluxEngine.cydsn/main.c
@@ -21,6 +21,10 @@ static bool drive1_present;
 
 static volatile uint32_t clock = 0; /* ms */
 static volatile bool index_irq = false;
+/* Duration in ms. 0 causes every pulse to be an index pulse. Durations since
+ * last pulse greater than this value imply sector pulse. Otherwise is an index
+ * pulse. */
+static volatile uint32_t hardsec_index_threshold = 0;
 
 static bool motor_on = false;
 static uint32_t motor_on_time = 0;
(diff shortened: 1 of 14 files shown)


davidgiven:
Belatedly (sorry, Christmas): Appveyor broke, which is why the check has failed. I've updated the configuration. Could you merge from master and repush, please? Thanks!

davidgiven:
BTW, it _should_ be possible to autodetect the number of hard sectors --- if there's more than one index pulse within, say, 100ms then it's clearly a hard-sectored disk; the interval to the index hole ought to be half the size of the others... if we start the disk running and track the minimum and maximum intervals plus frequencies, then determining the various numbers ought to be reasonably simple. WDYT?

ejona86:
> it should be possible to autodetect the number of hard sectors --- if there's more than one index pulse within, say, 100ms then it's clearly a hard-sectored disk; the interval to the index hole ought to be half the size of the others

This is actually how I started. Although then when I was looking at returning the sector information in the speed_frame and plumbing it through the command line tool I thought it was going to cause trouble. Looking at it again, it'll probably be fine. I'll try it out (Thursday or this weekend).

I closed/reopened the PR to retrigger AppVeyor.

I saw in https://github.com/davidgiven/fluxengine/pull/205 that you committed FluxEngine.hex. Should I do the same?

Also, while I increased the protocol version, I noticed that if the code zero-extended structures this would not be a breaking protocol change. If that sounds good to you, I'd be happy to clear out unused parts of structures to avoid bumping the protocol version.

davidgiven:
Hmm. Yes, you're right, let's keep it with the manual setting --- it's not like anyone's going to use hard-sectored disks by accident.

Regarding checking in FluxEngine.hex: yes, please. However, it's import to make sure you're up to date with master first, otherwise the changes there will be lost. Unfortunately it's not possible to build the firmware on the CI system so it always has to be built manually.

Thank you for doing this! It's something I've been wanting for ages...

BTW, The Verilog sequencer is supposed to have code in it where if there's a index pulse in the incoming bitstream it stops and waits for an index pulse on the hardware, so as to support writing to hard-sectored disks, but that's not done yet (the Verilog has gone through a lot of iterations as people have pointed out the bugs).

davidgiven:
Merged. Thanks!

Issue #201, opened 2020-12-27
Unable to write 10 sectors per track Atari ST disks


keithclark:
Hi. I'm built a flux engine so I can archive and recover data from my ageing Atari ST and Amiga disk collections. At the moment I'm focusing on the Atari Stuff. The hardware/software works well and I've managed to image several different disks. What I'm struggling with is writing those images back to disk. Disks I've read with fluxengine load in an emulator but when I use fluxengine write these images back to a physical floppy disk, they don't work.

Disks are various formats (9-10 sectors per track, 80/82 tracks). Disks with executable boot sectors (demos) start loading so the first 512 bytes are writing correctly, but fail right after that. The stdout dumped when writing doesn't seem to make sense to me — I'm expecting to see a linear list of sectors being written to the disk but data seems out of order.

Am I missing something obvious?

davidgiven:
What command line arguments are you using when writing? FluxEngine is unable to detect the disk parameters, so you have to explicitly tell it (`--preset-720` or similar). The out-of-order sectors is actually normal; they're not stored in linear order on the disk. Look up 'interleave'.

It's interesting that the first sector loads. That suggests it's _nearly_ working. It might be a side or track spacing problem, but AFAICR the Atari ST uses pretty vanilla 720kB IBM scheme disks, the same as the PC.

keithclark:
My test file `ibm.img` is a 2 side, 80 track, 10 sector, 512 byte per sector disk image, which was extracted by flux engine.

I've tried:

```
fluxengine write ibm --ibm-preset-720 -i ibm.img
```

Flux engine logs the following to the terminal:

```
reading 80 tracks, 2 heads, 9 sectors, 512 bytes per sector, 720 kB total
Writing to: :d=0:s=0-1:t=0-79
```

This is obviously wrong as the source image is 10 sectors per track. The boot sector executes but the FAT is missing entries, which I assume is because those entries are on sector 10. The BIOS Parameter Block of the disk still reports 10 sectors per track.

Next I tried specifying the input format:
```
fluxengine write ibm --ibm-preset-720 -i ibm.img:c=80:h=2:s=10:b=512
```

This produces what I was expecting:

```
reading 80 tracks, 2 heads, 10 sectors, 512 bytes per sector, 800 kB total
Writing to: :d=0:s=0-1:t=0-79
```

The missing files now appear on the disk but it still won't run. The Atari reports it is unable to read the disk. This happens for any disk I attempt to write.

[...]

PR #203, opened 2021-01-04
Write Atari ST formats

keithclark:
This PR addresses #201.

It adds a new write option for creating Atari ST disks and registers the standard `.st` file extension so disk images can be used without having to rename them first. The standard IBM writer is used but many of the options are preset for ST disks (no IAM, fixed clock rates / speeds etc.).

Atari ST disks can be written using:

```
fluxengine write atarist --st-preset-820 -i my-disk-image.st 
```

The following format presets are available:

flag | format
-|-
--st-preset-360 | 360kB disk (1 side, 80 tracks, 9 sectors)
--st-preset-370 | 370kB disk (1 side, 82 tracks, 9 sectors)
--st-preset-400 | 400kB disk (1 side, 80 Tracks, 10 sectors)
--st-preset-410 | 410kB disk (1 side, 82 tracks, 10 sectors)
--st-preset-720 | 720kB disk (2 sides, 80 tracks, 9 sectors)
--st-preset-740 | 740kB disk (2 sides, 82 tracks, 9 sectors)
--st-preset-800 | 800kB disk (2 sides, 80 tracks, 10 sectors)
--st-preset-820 | 820kB disk (2 sides, 82 tracks, 10 sectors)

### DIFF
--- a/doc/disk-atarist.md
+++ b/doc/disk-atarist.md
@@ -1,0 +1,73 @@
+Disk: Atari ST
+==============
+
+Atari ST disks are standard MFM encoded IBM scheme disks without an IAM header.
+Disks are typically formatted 512 bytes per sector with between 9-10 (sometimes
+11!) sectors per track and 80-82 tracks per side.
+
+
+Reading disks
+-------------
+
+Just do:
+
+    fluxengine read atarist
+
+...and you'll end up with an `atarist.st` file. The size of the disk image will
+vary depending on the format.
+
+Writing disks
+-------------
+
+FluxEngine can also write Atari ST scheme disks.
+
+The syntax is:
+
+    fluxengine write atarist -i input.st <options>
+
+The format of `input.st` will vary depending on the kind of disk you're writing,
+which is configured by the options. By default FluxEngine will write an 80
+track, 9 sector, double-sided disk. If that doesn't match your target format you
+will need to pass some options. There are some presets, which you will almost
+certainly want to use if possible:
+
+  - `--st-preset-360`: a 360kB 3.5" disk, with 80 cylinders,
+  1 side, and 9 sectors per track.
+  - `--st-preset-370`: a 370kB 3.5" disk, with 82 cylinders,
+  1 side, and 9 sectors per track.
+  - `--st-preset-400`: a 400kB 3.5" disk, with 80 cylinders,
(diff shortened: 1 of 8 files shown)


davidgiven:
Thank you very much for this!

Could you merge from master and push again, please? Appveryor changed their configuration and I had to fix it.

keithclark:
Fixed.

keithclark:
Would be nice to document this here: https://github.com/davidgiven/fluxengine/tree/master/doc. Can do that in a separate PR if needed?

keithclark:
For consistency, would it make sense to add `fluxengine read atarist`? Even it it's just a wrapper for reading IBM disks?

davidgiven:
Sorry, been stuck with other things.

Yes, I think a `read atarist` command would be a good idea --- even if all it does is write to `atarist.st` instead of `ibm.img`. Documentation would be fine, too, but I'm happy to do that.

keithclark:
I've added an Atari ST reader front-end and a documentation page. I think that covers off most common uses-cases for reading/writing Atari ST formatted disks.

davidgiven:
Sorry about the delay --- I've actually just fixed a firmware bug which was causing pulse timing to be slightly wrong when writing images.

Can I ask you to update from trunk again, flash the new firmware, and check that the Atari disks can be written and read back again? I think the bug was only affecting 1440kB IBM formats, but the effect of it is that tracks are now slightly shorter than they were, so if your sector timings are marginal there's a chance the last sector will be truncated at the end of the track (as was happening with the 1440kB format --- `--ibm-gap3-bytes` was too big).

keithclark:
I've rebased my changes onto yours and flashed the firmware to the latest version. Disk writing works but I am seeing intermittent failures when trying to load disks into my Atari ST - some are hanging at the boot sector. If I rewrite the disk it will work fine. I'm not sure if this is related to your firmware update or if its just a coincidence.

I'll perform some more tests when I get a chance and let you know how they go. 

I think the frontend code is fine though, it's doing what I expect it to.

davidgiven:
Hmm. Could be write failure --- writing a disk doesn't do any verification (it's on my list of things to do), so you'll only know if it failed by trying to read things back again. It's a possibility it's just normal spurious errors, which happen occasinally. I'll go ahead and merge.

PR #214, opened 2021-01-17
added parameter swapSides to IBM encoder

hpingel:
According to https://github.com/Distrotech/fdutils/blob/master/src/mediaprm there are some obscure disk drives out there that expect the sides 0 and 1 to be swapped (search for "swapsides"). We are talking about

- Commodore 1581 3.5" Disk Drive
- CMD FD-2000 (and FD-4000)
- Thomson TO7

Remark: I know nothing about the last one and so I don't know if it uses IBM encoding at all. But the Commodore related drives above do use IBM encoding

So for those special cases the IBM writer needs another parameter to deal with them, I came up with "swapSides" (bool).

The big question is: Is this now cumbersome for all those "ordinary" cases where presets have to provide a false value for swapSides - like for example the Atari ST presets in the other Pull Request.

### DIFF
--- a/arch/ibm/encoder.cc
+++ b/arch/ibm/encoder.cc
@@ -102,6 +102,8 @@ static uint8_t decodeUint16(uint16_t raw)
 std::unique_ptr<Fluxmap> IbmEncoder::encode(
 	int physicalTrack, int physicalSide, const SectorSet& allSectors)
 {
+	if (_parameters.swapSides)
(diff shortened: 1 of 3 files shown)


davidgiven:
LGTM --- I'll update and merge this. You're right about the Atari ST presets; I wonder whether I should combine the Atari ST command with the generic IBM one.

Issue #107, opened 2019-09-14
Report: Reading of CMD FD-2000 HD disks successful


hpingel:
Hi David,

This is not so much an issue, just some feedback. As I don't own a FD-2000 disk drive from Creative Micro Designs (CMD) I have written a d2m image file (downloaded from the internet) to a 3.5" HD disk using my ancient Linux box tweaking the disk drive in there as follows:

```
mknod /dev/fd0cmdfd2m b 2 124
setfdprm /dev/fd0cmdfd2m HD DS sect=10 ssize=1024 cyl=81 swapsides

time fdformat /dev/fd0cmdfd2m
time dd if=Downloads/dummy.fd2 of=/dev/fd0cmdfd2m count=1620 bs=1024
```
The FD-2000 HD disk format (a DD format can also be written by the drive) seems to have 81 tracks and 10 sectors of 1024 bytes per track. It stores 1620 KB of end-user data.

> CMD FD2000 disk drive, a third party floppy disk drive for the
> Commodore line of computers with double of the capacity:
> "CMDFD2M":
>  DS HD sect=10 cyl=81 ssize=1024 fmt_gap=100 gap=12 swapsides
> 

As it wants to be downward compatible to the Commodore 1581 drive, it also swaps the sides (=swapsides).

FluxEngine parmeters used:
`fluxengine read ibm --revolutions=2 -s ":d=0:t=0-80:s=0-1"`

Ressources:
https://web.archive.org/web/20060622053148/http://www.cmdweb.de/fd.htm
https://www.c64-wiki.de/wiki/FD-2000
http://www.unusedino.de/ec64/technical/formats/d2m-dnp.html
https://cbm8bit.com/articles/user-contributions/howto_d1m_d2m_d4m
Visualisation:
![fd2000](https://user-images.githubusercontent.com/646865/64913938-b02f8700-d749-11e9-9391-16ca08974cc5.png)

The only thing missing from FluxEngine is support for the image file extension "d2m" or even "d4m" for the CMD FS-4000.

davidgiven:
(belatedly) Good news --- thanks for the report.

PR #218, opened 2021-01-18
Adding preset for Commodore 1581 (cbm1581) to IBM writer frontend

hpingel:
Hi,

Now this change is just a proposal - in two ways:

First, my preset for Commodore 1581 might not fit in well in this file - maybe it should get its own source file. (Especially as in a couple of days/weeks I will come up with a preset for CMD FD-2000 once I have found somebody who owns one and can confirm that the preset is ok. Some details regarding FD-2000 are in #107)

Secondly, the gap values of the 1581 preset are good enough for me and my tests with a real 1581 drive but might be matter of change once somebody reviews them / looks at diagrams, etc.

So this is just the current state of things and if you want - merge it, if not, it's ok for me - I can play with my 1581 images locally anyway.

### DIFF
--- a/src/fe-writeibm.cc
+++ b/src/fe-writeibm.cc
@@ -95,6 +95,7 @@ static ActionFlag preset1440(
 		gap2.setDefaultValue(22);
 		gap3.setDefaultValue(80);
 		sectorSkew.setDefaultValue("0123456789abcdefgh");
+		swapSides.setDefaultValue(false);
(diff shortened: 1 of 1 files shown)


davidgiven:
Sorry for the delay --- working on other projects. Yup, LGTM.

davidgiven:
Merged. Thanks!

PR #221, opened 2021-01-21
Enforce usage of Greaseweazle firmware v0.24 after adapting CMD_READ_…

hpingel:
…FLUX and CMD_WRITE_FLUX

### DIFF
--- a/lib/usb/greaseweazle.h
+++ b/lib/usb/greaseweazle.h
@@ -7,7 +7,7 @@
 #define EP_OUT 0x02
 #define EP_IN 0x83
 
-#define GREASEWEAZLE_VERSION 22
+#define GREASEWEAZLE_VERSION 24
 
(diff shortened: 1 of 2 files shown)


davidgiven:
Sorry about the delay --- been working on other things...

Thanks very much for this! However I would rather like to keep compatibility with v23 as well as v24, as given that I'm a third-party client I don't really want to demand that people upgrade their Greaseweazle firmware. Any chance of selecting the appropriate API based on the reported version? It should be a simple matter of stashing the version number in a global variable in a member variable and then using one or the other based on the value.

hpingel:
Thanks. Yeah, I will look into a flexible approach to allow v0.22+. May take some time!

davidgiven:
I've actually overhauled all the USB stuff (and, sadly, had to take out the GreaseWeazle autodetection because I can't make it work on Windows), and merged your code along with it. Thanks very much!

PR #222, opened 2021-01-23
Add support for North Star hard-sectored floppies.

hharte:
North Star Floppy disks use 10-sector hard sectored disks with either FM
or MFM encoding.  They may be single- or double-sided.  Each of the 10
sectors contains 256 (FM) or 512 (MFM) bytes of data.  The disk has 35
cylinders, with tracks 0-34 on side 0, and tracks 35-69 on side 1.
Tracks on side 1 are numbered "backwards" in that track 35 corresponds
to cylinder 34, side 1, and track 69 corresponds to cylinder 0, side 1.

The North Star sector format does not include any head positioning
information.  As such, reads from North Star floppies need to by
synchronized with the index pulse, in order to properly identify the
sector being read.  While there is a command line option:
--sync-with-index, that does this, the North Star reader forces this
behavior by default.

Due to the nature of the track ordering on side 1, an .nsi image writer
is provided for double-sided disks.  The .nsi image writer supports both
single- and double-sided disks, but single-sided disks may also be
written in .img format.

### DIFF
--- a/FluxEngine.cydsn/main.c
+++ b/FluxEngine.cydsn/main.c
@@ -79,6 +79,7 @@ CY_ISR(index_irq_cb)
      * the track. */
     static bool hardsec_index_irq_primed = false;
     static uint32_t hardsec_last_pulse_time = 0;
+    uint32_t index_pulse_duration = clock - hardsec_last_pulse_time;
 
     if (!hardsec_index_threshold)
     {
@@ -87,12 +88,18 @@ CY_ISR(index_irq_cb)
     }
     else
     {
-        index_irq = hardsec_index_irq_primed;
+        /* It's only an index pulse if the previous pulse is less than
+         * the threshold.
+         */
+        index_irq = (index_pulse_duration <= hardsec_index_threshold) ?
+            hardsec_index_irq_primed : false;
+
         if (index_irq)
             hardsec_index_irq_primed = false;
         else
             hardsec_index_irq_primed =
-                clock - hardsec_last_pulse_time <= hardsec_index_threshold;
+                index_pulse_duration <= hardsec_index_threshold;
+
         hardsec_last_pulse_time = clock;
     }
     
@@ -275,7 +282,6 @@ static void seek_to(int track)
         CyWdtClear();
     }
     CyDelay(STEP_SETTLING_TIME);
-    
     TK43_REG_Write(track < 43); /* high if 0..42, low if 43 or up */
     print("finished seek");
 }
@@ -305,7 +311,7 @@ static void cmd_measure_speed(struct measurespeed_frame* f)
     while (!index_irq)
     {
         elapsed = clock - start_clock;
-        if (elapsed > 1000)
+        if (elapsed > 1500)
         {
             elapsed = 0;
             break;
@@ -416,7 +422,6 @@ static void cmd_read(struct read_frame* f)
     seek_to(current_track);
     SIDE_REG_Write(f->side);
     STEP_REG_Write(f->side); /* for drives which multiplex SIDE and DIR */
-    
     /* Do slow setup *before* we go into the real-time bit. */
     
     {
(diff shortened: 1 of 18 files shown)


hharte:
FYI @davidgiven, @ejona86

hharte (commented):
I changed the hard-sector ID from using a counting method to using a direct calculation, based on the elapsed time since the index pulse.  This has proven to work better with problematic disks.

hharte:
This should be good to go for reading North Star hard-sectored floppies using a standard 48TPI 300RPM (360K) drive.  I was to successfully read all of the North Star floppies I had.  Thank you for making such a great device!

FluxEngine North Star

davidgiven:
Apparently I completely failed to notice this PR. Sorry! And, that external enclosure looks awesome! They made _the_ best power switches. That's a IBM 360kB 48tpi drive in it, by the star on the front; I'm looking for one of these but have completely failed to find one...

I'm currently in the process of reworking the entire architecture to make it more consistent and flexible. See the `protobuf` branch if you want a preview. This means that merging won't be trivial and some rework will need to be done, but I'll do that myself, eventually --- thanks very much for the PR.

This is also the first format which has a non-trivial mapping between track in the sector image and cylinder on the disk, and I'm wondering if this is quite correct in the core code. The primary key for a SectorSet is supposed to be the _logical_ track, where the conversion from logical to physical happens in the encoder/decider. I see that you're doing it in the .nsi reader/writer. This is incorrect, as it ends up with the sector's logicalTrack/logicalSide being set incorrectly, but I'm not sure whether it'll actually work when doing it 'properly'. This is also rather bound up with the 40-track support, which is problematic and probably needs throwing away.

It might actually be worth adding options to the .img image reader/writer to allow for image formats like this. Acorn images tend to be HCS rather than CHS too.

hharte:
Thanks David,

Agree with you on the enclosure and power switches!  All IBM floppies with the star are 48-TPI, but not all IBM 48-TPI floppies have the star.  I have one newer one that is 48-TPI without the star.  They also switched drive manufacturers at some point.  Those IBM external floppy drives show up on eBay from time to time.  Just remove the factory 37-pin cable, make a short 34-34-pin cable, and the FluxEngine fits right inside.  I use a USB "extension" cable so I can have cables coming out the back for normal use as well as firmware update without needing to open the case.  One thing to do though is to adjust the drive RPM to exactly 300 RPM while the case is open, it makes the writing work much more reliably.

I'm fine with the approach of adding different track ordering to the .img file reader/writer.  That would reduce a lot of code duplication, at the expense of some complexity.

The firmware change that is included in this pull request may be good to take regardless of the North Star support.  The firmware change fixes a race condition regarding counting the hard secor pulses.  It is useful for Micropolis (Vector Graphic) and probably others.  The Micropolis reading works ok, because the cyl/sector information is encoded in the sector header.  Writing Micropolis will be problematic, as it's possible to write the sectors incorrectly with the current firmware.  For Norh Star, there is no positioning data encoded in the sector header, so correct counting of the index pulses is absolutely critical for both reading and writing.

Take care,
Howard

davidgiven:
I've merged the Northstar code into #266, migrating it to the new architecture. It's entirely possible I've mangled the image reader and writer. PTAL?

I need to rework the way sectors are mapped to images and vice versa. With the new much more flexible configuration system, I'm thinking of adding a new `geometry` configuration paragraph; there'd then be a separate stage between the encoder/decoder and image reader/writer which maps between filesystem blocks and sectors. The image reader/writer then doesn't have to know about tracks and sectors. This should, I hope, allow the `.img` reader/writer with an appropriate Northstar geometry mapper. It needs thought, though.

ejona86:
BTW: a while back I did review the hard-sector detection code and it seemed good. `&&` vs `? : false` may be more clear, but whatever.

I've believed for a while that the CP/M images with the Micropolis are in "backwards" on the second side, but I've not gone and verified it so I hadn't sent any documentation updates.

davidgiven:
That is now merged and will go live as soon as the autobuilder finishes. Thanks very much!

PR #229, opened 2021-02-07
Add support for slow drives and 40-track disks in 80-track drives.

hharte:
Some drives (like the Micropolis 1015-II require a step rate of 30ms, others, like 360K PC drives specify an 8ms step rate.  Add the ability to specify the step rate and settle time on command-line parameters.  Also allow the drive to be double-stepped (handy for reading 40-track disks in 80-track drives.)

### DIFF
--- a/FluxEngine.cydsn/main.c
+++ b/FluxEngine.cydsn/main.c
@@ -7,8 +7,8 @@
 #include "../protocol.h"
 
 #define MOTOR_ON_TIME 5000 /* milliseconds */
-#define STEP_INTERVAL_TIME 6 /* ms */
-#define STEP_SETTLING_TIME 50 /* ms */
+#define STEP_INTERVAL_TIME_DEFAULT 6 /* ms */
+#define STEP_SETTLING_TIME_DEFAULT 50 /* ms */
 
 #define DISKSTATUS_WPT    1
 #define DISKSTATUS_DSKCHG 2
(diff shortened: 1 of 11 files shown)


davidgiven:
Sorry for the delay --- I've been working on other projects.

Thanks for this, but actually there already is support for double-stepping! Do something like `-s t=0-79x2`. I would like to keep the physical track numbering matching that in 96tpi drives if possible, so for 40-track drives really I want the _opposite_ conversion to the one you've done here (where when FluxEngine seeks to physical track 22, the drive actually goes to track 11).

The step and settling time would be handy.

hharte:
Hello David,

thank you for the feedback.  I agree the x2 does work for reading.  For writing, it presents an issue, because the .img source file is also “double stepped” so when writing a 360K floppy for example, one would need a 720K .img with every other track padded with filler.  If you know of a way around this, please let me know.

davidgiven:
Sorry again for the very long delay... I have just added formal support for 40-track disks and drives. It's not very clean but should work; images can be read with `:t=2` to control the logical-to-physical mapping, and there's a `--40-track` option to tell FluxEngine that a 40-track drive is attached so it needs to halve track numbers. FluxEngine uses 80-track track numbers internally throughout (so the tracks on a 40-track drive are 0, 2, 4, 6 etc). See the sections on logical vs physical tracks and 40-track drives at http://cowlark.com/fluxengine/doc/using.html.

hharte:
Thanks David.  I'd still like to add the ability to change the head step rate and settle time.  I'll close this PR and submit a new one with just those changes.

PR #241, opened 2021-04-12
Update fe-writeibm.cc

wybren1971:
Added a HP-LIF preset to write HP-LIF floppy disks for old HP analyzers in need of LIF formatted disks with 76 tracks and sectors of 1024 bytes.


davidgiven:
That's great --- thanks!

PR #248, opened 2021-04-30
Writing the disk based on the geometry in the IMD image

wybren1971:
Hallo David,

As i explained earlier i couldn't set the geometry to write a disk based on the geometry in the IMD image. I managed to get it working with static class members in the IbmEncoder. I submit my changes to you for review and to merge into you're code.
I also added the missing functionality for the optional cylinder map and optional head map. and made sure the sector status was correctly set. (further some small fixes to the code)

I really enjoy fluxengine and hope we can make this even better.

Greetings Wybren

### DIFF
--- a/lib/decoders/decoders.proto
+++ b/lib/decoders/decoders.proto
@@ -19,7 +19,7 @@ import "arch/zilogmcz/zilogmcz.proto";
 import "lib/fluxsink/fluxsink.proto";
 import "lib/common.proto";
 
-//NEXT: 29
+//NEXT: 30
 message DecoderProto {
 	optional double pulse_debounce_threshold = 1 [default = 0.30,
 		(help) = "ignore pulses with intervals shorter than this, in fractions of a clock"];
@@ -61,5 +61,7 @@ message DecoderProto {
 		(help) = "how many times to retry each track in the event of a read failure"];
 	optional string write_csv_to = 23
 		[(help) = "if set, write a CSV report of the disk state"];
+	optional bool skip_unnecessary_tracks = 29 [default = true,
+		(help) = "don't read tracks if we already have all necessary sectors"];
 }
 
--- a/lib/readerwriter.cc
+++ b/lib/readerwriter.cc
@@ -27,6 +27,12 @@ enum ReadResult
     BAD_AND_CAN_NOT_RETRY
 };
 
+enum BadSectorsState
+{
+	HAS_NO_BAD_SECTORS,
+	HAS_BAD_SECTORS
+};
+
 /* In order to allow rereads in file-based flux sources, we need to persist the
  * FluxSourceIterator (as that's where the state for which read to return is
  * held). This class handles that. */
@@ -108,8 +114,7 @@ static std::set<std::shared_ptr<const Sector>> collectSectors(
     return sector_set;
 }
 
-/* Returns true if the result contains bad sectors. */
-bool combineRecordAndSectors(
+BadSectorsState combineRecordAndSectors(
     TrackFlux& trackFlux, AbstractDecoder& decoder, const Location& location)
 {
     std::set<std::shared_ptr<const Sector>> track_sectors;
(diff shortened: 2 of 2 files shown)


wybren1971:
Hello David,

I completed the writing of the code for archiving disks in the IMD format. A comment can be included in the IMD file.
IMD works now for disks with de ibm filesystem. The sector-skew has to be the same for the whole disk otherwise fluxengine won't be able to write the file to disk. i updated the documentation for de IMD format. Have a look at it and let me know what you think. i also ironed out some bugs still in the code. Is there a better way to know if FM or MFM coding is used? besides the boolean i added to the class?

Fluxengine is flexibel when reading disks, but when writing it is not that flexible. For example the sector-size and sector-skew have to be the same for the whole disk. Why is that? When reading a disk, sector-size and sector-skew can change and fluxengine has no problems reading and storing it.

Hoping to hear from you.

Kind regards,

Wybren

davidgiven:
Excellent, thanks --- I'll take a look.

Regarding writing and sector-size and -skew: this works when reading because the decoder will autodetect most parameters on a per-track level. Skew's ignored as it reads all sectors and then reassembles them, regardless of where they are. However, when writing, these parameters can't be autodetected and have to specified explicitly. Currently the only reason for having a single set is that nobody's really asked for anything else! It's actually the responsibility of the encoder to take these parameters into account, so it can do whatever it likes, but the parameters are all in flags and so can't change.

Overhauling all this wouldn't be hard, probably by having the encoder take a pointer to a routine that creates the parameters so that they can be varied on the fly; the hardest part would be to find a way to let the user specify them. Many CP/M machines use multiformat disks, so that'd be useful.

Currently the workaround is to write the disk in multiple passes, each one with different parameters and writing a different set of tracks.

wybren1971:
Just detected a small bug with HD diskettes. i set the sectorskew with integers. so 0 to 18. but fluxengine wants 0-9 a-g.
and fixed some bad-checksum errors. (I had a disk with bad sectors, hadn't had that before). well it works her, but i haven't tested all the possibilities... 
I added the possibility to add a comment to the IMD file for archive purposes. But i don't know if it fits with fluxengine.
just test it out and let me know what you find.

I understand what you say regarding the decoder. But i find C++ still a bit difficult. so i'm still learning. If you have any coding suggestions ... i am always open for suggestions..

wybren1971:
Did the last few bugfixes and clean up in messages. this version works at all the disks i have thrown at it. Maybe you can test some as well. let me know:-)
It would be nice if fluxengine can write a disk based on the specifications in a diskimage. This works now with IMD except with the limitation of the sector-skew an sector-size. Another limitation is that it works now only for IBM formatted disks. While it should also work for amiga disks etc.
Thanks for your great work. i learn a lot from reading your code.

Kind regards Wybren

wybren1971:
i keep tinkering... But now it is functioning good enough. I can work with this for my own collection of old disks. If you decide to take it into the base code of fluxengine, i'm willing to provide assistance with correcting the code in case of a bug reported by a user. Or help investigate.
Just let me know.

davidgiven:
Thanks very much. This has actually spurred me to do the much-delayed reworking of the configuration system, which with luck should make this much easier than using static global variables. Once that's done I'll merge this.

wybren1971:
Yeah, i follow the modifications you make in protobuf. Looks promising. IMD contains disk information and can then set the parameters for writing directly which should make everything more user friendly. I'm looking into making a wrapper around fluxengine with QTCreator. So fluxengine gets easier to use for people who don't like commandline interfaces... I'm looking forward to the next release of fluxengine😁

wybren1971:
i updated the IMD read and write and made them compatible to protobuf (commit 5b3a99f) . One issue still remains to be solved. Maybe you got some ideas.. in imdimagewriter.cc i have to write to the IMD image the kind of encoding used FM or MFM. Where can i get this setting do i can write it to the IMD image... In this commit i use always MFM...

wybren1971:
Can i do something to help? i made the code as compatible as i could. One issue is how to determine the FM or MFM encoding needed to write to the IMD file. see also my comment on 26 of may.
Hope to hear from you soon.

Greetings Wybren

davidgiven:
Sorry for the very long delay... after much refactoring we should now be in a position to actually do this. There's actual infrastructure in place to support it. See https://github.com/davidgiven/fluxengine/blob/master/lib/imagereader/d88imagereader.cc#L64 for an example. I did have a look at converting the IMD reader but it turns out to be rather more complicated than I first expected and you know more about the format than I do. Are you still interested in this?

wybren1971:
Hello David,

Yes i am still interested in this and i am not alone. I will see if i can make head or tales of the code and submit you a pull request based on the latest version of fluxengine. looks like you made a lot of progress!

PR #257, opened 2021-05-11
Commodore 64 encoder

wybren1971:
Hello David,

I had the need to write some D64 images back to floppy so i wrote this encoder. And it works great! Just wrote a couple d64 images back to disk and read it with my old commodore 64 and 1541 II floppy drive.

I share the code with you if you would like to merge it into your code.

Kind regards

Wybren


wybren1971:
I even the disk ID back needed to write in the header from the data in de d64 file. The Disk ID was stored in the BAM. Some C64 check this. No they are written (in the correct order;-)

davidgiven:
That's brilliant, thank-you. I'll merge this once I'm done with the current huge rewrite. Don't worry about the git issues, I'll fix things myself as I'll need to adapt your code to the new configuration system.

davidgiven:
Hello --- I've pulled out the encoder and d64 reader and converted them to the new architecture; they're in the wybren1971-Commodore-64-encoder branch. Could you take a look, please? The format name has changed, so to do a write you'll need `fluxengine write commodore1541 ...`.

Also, I have a suspicion that the D64 reader and writer is no longer needed as the .img reader and writer can be configured to do the same job. It's been a while; can you confirm that the D64 format is just all the sectors one after the other, with no header?

wybren1971:
Hello David,

I will take a look today at the new branch. And test it with my old commodore.

You are correct that the .d64 image just contains the 256 byte sectors in a row and no header information. I had to reassemble the header information from sources of the internet because it wasn't in the d64 image. Maybe adding the g64 image is also a feature option as it contains more disk information. All be it for purely preservation reasons, because you can't hook it up to vice.

My fluxengine GUI is coming along and I will share a first beta with you soon. I shamelessly used you're flying cow image in it....

I'll get back to you...

wybren1971:
Hello David,

After downloading the sources from the new branch, i compiled these. I got an error "Makefile:26: *** You must have these pkg-config packages installed: zlib sqlite3 libusb-1.0 protobuf.  Stop." 
So i installed the latest release from protobuf from here and followed the [readme](https://github.com/protocolbuffers/protobuf/blob/master/src/README.md). After protobuf was installed. i gave a make in the fluexengine directory and i was given the following output:

> make
[1/1] LINK .obj/dbg/libfrontend.a .obj/dbg/libbackend.a .obj/dbg/libproto.a .obj/dbg/libfmt.a .obj/dbg/libagg.a
FAILED: fluxengine-debug 
g++ -pthread  -o fluxengine-debug .obj/dbg/libfrontend.a .obj/dbg/libbackend.a .obj/dbg/libproto.a .obj/dbg/libfmt.a .obj/dbg/libagg.a -O0 -g -L/usr/local/lib -lz -lsqlite3 -lusb-1.0 -lprotobuf -lpthread
.obj/dbg/libfrontend.a(fe-fluxtoau.o): In function `mainConvertFluxToAu(int, char const**)':
/home/wybren/Documents/fluxengine/src/fe-fluxtoau.cc:35: undefined reference to `FlagGroup::parseFlags(int, char const**)'
/home/wybren/Documents/fluxengine/src/fe-fluxtoau.cc:44: undefined reference to `setHardwareFluxSourceDensity(bool)'
/home/wybren/Documents/fluxengine/src/fe-fluxtoau.cc:45: undefined reference to `FluxSource::create(FluxSpec const&)'
.obj/dbg/libfrontend.a(fe-fluxtoau.o): In function `__static_initialization_and_destruction_0(int, int)':
/home/wybren/Documents/fluxengine/src/fe-fluxtoau.cc:13: undefined reference to `FlagGroup::FlagGroup(std::initializer_list<FlagGroup*>)'
.obj/dbg/libfrontend.a(fe-fluxtoau.o):(.data.rel.ro.._111+0x0): undefined reference to `hardwareFluxSourceFlags'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
Makefile:45: recipe for target 'all' failed
make: *** [all] Error 1

But fluxengine worked with the new commandline interface, so i ignored this error. I took one of my commodore1541 diskettes and read them with fluxengine:
`./fluxengine read commodore1541 -s drive:0 -o uridium.d64 --copy-flux-to uridium.flux -c 0-39 -h 0`
and i got an image which i tested with Vice and worked.
Then i got an empty 5.25 inch floppy and wrote the image back to it with fluxengine:
`./fluxengine write commodore1541 -i uridium.d64 -d drive:0 -c 0-39 -h 0`
All went well. And when i put this new written disk in my old 1541 II diskdrive attached to my commodore 64 this would load and run!.
So all seems to be in order and a unicorn can be added at the homepage;-)

In short. It works for me and i can confirm that the D64 format is just all the sectors one after the other, with no header!

davidgiven:
I've just done the merge; the autobuilder should do a release soon...

This PR should now be obsolete; I'll tackle the IMD code next. Thanks very much!

hpingel:
Hi @wybren1971 and @davidgiven,

Thanks @wybren1971 for your work on the commodore1541 write mode! I have tested this the other night and so far I was not successful to create a valid disk that would be readable (in either real 1541 drives or the PC drive I'm using with Fluxengine).

So it might be possible that my PC drive has a faulty write head, that could be one possibility. But I also wanted to clarify: @wybren1971, are you using a 40-track-only PC drive with Fluxengine?

My concern is that there might be something wrong at the moment when writing a 40 track d64 image to a floppy disk using an 80 track capable PC drive. The drive I use is an EPSON SD-680L (but I would also have a TEAC to test with). But maybe I'm thinking wrongly - Fluxengine will always expect the 5.25" PC drive to be capable of writing 80 tracks? Hmmm.

wybren1971:
Hi @hpingel,

Thanks for testing the commodore1541 encoder. Sorry to hear it didn't result in a working floppy.
To clarify things, i use a 5.25" 40 track drive and that's the only drive i tested it with. I used that one because the 1541 is also a 40 track drive. i have a 5.25" 80 track drive, i could test with that if you like.
I assume you used double step mode (79x2) to write the disk?  So the disk always skips one track?
and that de floppy you used is a 40 track commodore1541 floppy?

otherwise i don't know an option... i know that writing a 40 track floppy with a 80 track drive and reading that in an other 40 track floppy drive is tricky..
I will try my 80 track drive to write a 40 track floppy and let you know my experience with that.

i'll get back to you.

Wybren

hpingel:
Thank you for your quick reply, @wybren1971!
I would be interested to learn how your test goes! The is also the clockrate issue - I think my PC floppy drive rotates at 360 rpm per default and the 1541 rotates at 300 if I'm not mistaken.

For my tests I used a known-good C64 disk that is readable in Fluxengine once it was formatted or filled using a real 1541 II.
I only used the A side of the floppy disk as my PC drive is not modded with any index hole sensor mod.

As the parameters in Fluxengine were changed recently I was a bit confused as the old parameter system didn't work any more. So instead of using 0-79x2 (of which I don't know if it works only for reading or also for writing disks) I hacked the source code to distinguish between logical and physical tracks and I wrote only every second track. Maybe it would make more sense to write on all 80 tracks with duplicating each of the 40 tracks.

wybren1971:
Hello @hpingel ,

i tried today to write a 40 track commodore1541 disk in a 80 track drive but fluxengine crashed. i made an issue and i hope @davidgiven will look at it and maybe give some pointers on how to use the new protoconfig for writing 40 track floppies in a 80 track drive. when writing a commodore1541 disk in a 80 track drive.
Sorry i couldn't be of more help.
Reading a 40 track floppy in a 80 track drive is no problem. Vice could load from the resulting .d64 fine.

i have a spare 40 track 5.25" drive if you're interested in it;-)

Gandalf-ND:
> "The issue between 40- and 80-track drives is that the 80-track drives use a narrower head and read/write a narrower track. Thus an 80-track drive will read a 40-track diskette produced by either type of drive with no problem. However, a 40-track drive will generally have great difficulty reading any sectors (or full tracks) written in 40-track format by an 80-track drive. (Many retries may eventually produce a successful read.)" from: [link](https://retrocomputing.stackexchange.com/questions/14319/mismatched-floppy-disk-drive-reading-and-writing-what-should-one-keep-in-mind)

Writing a blank floppy (not even formatted earlier) with an 80 track drive and reading it on a 40 track drive might work. But if the floppy have been used previously it is probably quite futile. The problem is that the 40 track drive have a read head that is covering two tracks on the 80 track drive so the signal is a mix of two tracks. Even writing two identical tracks is a problem. You need to match the two tracks with the flux changes at the same position and that's not within the precision of the system.

If the floppy have been used before it will have information written to it in form of flux changes. Either as two separate narrow tracks or one wide track. The erase head on an 80 track drive is too narrow to erase the adjacent track so the 40 track drive will read half old flux and the new track written on an 80 track drive.

The only way I can see that it will work is if you can use a media eraser box, a box with powerful magnets that can degauss bulk media. I've seen one in my life. It was used to bulk erase old data tapes to reuse them. It could take up to 14" reels.

davidgiven:
Yup; FluxEngine should be capable of magnetically erasing a track, so the way the 40-track writing is supposed to work is that it writes data to even tracks and erases odd tracks. However, this hasn't been tested in real life yet. It's possible the drives don't behave the way I think they do (I have no working 40-track hardware).

hpingel:
I tried to write even tracks and erase odd tracks by hacking a few lines into the sourcecode and there was no quick/easy success to report in doing that. Maybe I have to you use a track offset? So far I wrote 0,2,4,6,... and erased 1,3,5,7,... 

I have never compared raw flux data so far which would maybe be helpful here. I tried formatting the disk in the 1541 first and then let Fluxengine write over it but there was no instant gain from doing that.

Of course it would be close to magic if Fluxengine could solve this 30-40 year old (more or less well-known) problem but at the moment I'm sceptical. 

@wybren1971 - if you have a 40 track drive to spare I would be interested but I would give @davidgiven the priority if he would like one. These things are hard to find nowadays. It took me a long time to find and obtain my two 5.25" drives in 2019 at a reaonable price. And I guess it hasn't become easier since then.

hpingel:
Astonishingly, writing works now with an 80 track drive, see #288.

PR #259, opened 2021-05-13
add write support for various 5.25" formats

salfter:
I wanted to create a FreeDOS boot floppy for an old Pentium III motherboard I had recently salvaged, but had the Teac FD-55GFR I'd bought 30 years ago for my first x86 box and some double-density floppies purchased NIB from an eBay seller.  (My longer-term interest in the FluxEngine is interoperation with the Apple II, but I'm just getting started with it.)  I'm not sure if there's a command-line incantation that sets up all the options just right to write 5.25" floppies, but this PR adds some presets for the common formats.  Separate options handle writing 360K disks in either double-density or high-density drives, and I included a 720K "quad-density" option (80 double-density tracks per side written in a high-density drive) that I used for cheap backups back in the day.  

So far, I've tested writing both 360K and 720K formats in a high-density drive, and I've booted FreeDOS from the 360K disk when the drive was hooked up to another system.  I have a double-density drive on order from another eBay seller and should probably snag some high-density disks as well to test that, but I suspect you can tell by looking at the source if I have the options right.

### DIFF
--- a/src/fe-writeibm.cc
+++ b/src/fe-writeibm.cc
@@ -116,6 +116,54 @@ static ActionFlag preset720(
 		set_ibm_defaults();
 	});
 
+static ActionFlag preset1200(
+	{ "--ibm-preset-1200" },
+	"Preset parameters to a PC 5.25\" 1.2MB disk.",
+	[] {
+		setWriterDefaultInput(":c=80:h=2:s=15:b=512");
+		setWriterDefaultDest(":t=0-79");
+		trackLengthMs.setDefaultValue(167);
+		clockRateKhz.setDefaultValue(500);
+		sectorSkew.setDefaultValue("0123456789abcde");
+		set_ibm_defaults();
+	});
+
+static ActionFlag preset720_qd(
+	{ "--ibm-preset-720-qd" },
+	"Preset parameters to a PC 5.25\" 720kB disk in a high-density drive.",
+	[] {
+		setWriterDefaultInput(":c=80:h=2:s=9:b=512");
+		setWriterDefaultDest(":t=0-79");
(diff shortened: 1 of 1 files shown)


davidgiven:
That's awesome, thanks!

However I am currently working on a big rewrite of basically everything, so I'll hold of merging this until that's done to avoid conflicts (the way disk formats are defined is changing completely).

There is Apple II decode support, but Apple II disks are barking mad due to essentially being software-defined. I can't replicate any of the tricks with spiral tracks or partially magnetised areas. Simple AppleDOS disks should work fine, though.

There's also rather poor support for 48tpi drives, which is going to be reworked, so if you have any ideas how you think it _should_ work, please let me know.

salfter:
> There is Apple II decode support, but Apple II disks are barking mad due to essentially being software-defined. I can't replicate any of the tricks with spiral tracks or partially magnetised areas.

Some of those tricks may require more support than you can get from a standard floppy drive.  For instance, since the head-positioning stepper motor phases are under software control, you can power two phases at the same time to get a half-step, which amounts to quarter-track positioning that was used by some copy-protection schemes.  Drives that only expose step & direction controls will never be able to duplicate that; you'd probably need to attach an Apple II floppy drive directly and twiddle the appropriate pins.

Most of what I have is standard ProDOS floppies (and a few DOS 3.3 floppies); the only tricks I used were notching them to write the back side (nearly everybody did that) and formatting 38 tracks instead of the standard 35 to squeeze another 12K per side (not everybody did that, but ProDOS supported it without modification).

davidgiven:
I've done the rework and merged a lot of these --- the 48tpi stuff is still TODO, though. Thanks very much!

PR #267, opened 2021-05-25
Northstar: Fix configurations.

hharte:
Tested read/write with 350K, 175K, and 87K disks.


davidgiven:
_facepalm_

davidgiven:
Also, I didn't know you could issue pull requests for other pull requests!

hharte:
Thanks David, I didn’t know that either, but I couldn’t figure out how else to submit a patch with the change to the review.

Issue #288, opened 2021-07-03
Fluxengine crash when writing a 40 track Commodore1541 disk in a 80 track drive


wybren1971:
Hallo @davidgiven ,

today i was trying to write a 40track disk in a 80 track drive  See also: #257  @hpingel. I specified the following command: 'write commodore1541 -i `"/data/commodore1541.d64" --input.image.img.physical_offset=1 --input.image.img.physical_step=2 -d drive:0 -c 0-79 -h 0"` 
fluxengine crashed with the following message:
`/bin/bash: line 1:  5604 Segmentation fault      (core dumped) "/home/wybren/Documents/given/fluxengine/fluxengine" write commodore1541 -i "/data/commodore1541.d64" --input.image.img.physical_offset=1 --input.image.img.physical_step=2 -d drive:0 -c 0-79 -h 0`

How can i specify in protoconfig that one logical track does not map directly onto one physical track? with the previous version i could change this with `:o=1:t=2`: `o` specifies the offset, and `t` specifies the step. So, with this format, cylinder 1 in the image will be written to track 3 on the disk.
I thought that --input.image.img.physical_offset=1 --input.image.img.physical_step=2 would do the same and write a 40 track commodore1541 disk in an 80 track drive.
Could you please help?

thanks

Wybren

davidgiven:
Well, that's not good. That _should_ have worked. The 40-track stuff got rather mangled in the config rewrite; I've been hoping to come up with a cleaner way to do it (but haven't).

So, I've fixed the crash and the logical:physical mapping, so this ought to work now. Could you try the version from the c64 branch, please? Also, your command is wrong; by using `--input.image.img` you're resetting the image type from D64 back to IMG, so it won't be read correctly. (There could plausibly be a warning there.) You just need this:

```
./fluxengine write commodore1541 -i image.d64 -d drive:0
```

wybren1971:
And this command works for both a 80 track drive and a 40 track drive?
i downloaded the branch, my fluxenginemachine is compiling....

wybren1971:
I tested this branch and it works fine!! on a 80 track drive.
What i tested:
 i used attached .d64 file (which i made reading from a old disk with the 80 track drive)
[commodore1541.zip](https://github.com/davidgiven/fluxengine/files/6758341/commodore1541.zip)
then i wrote the file back to the same disk (with the 80 track drive) and that disk loaded in my old 1541II drive and commodore 64.
[...]

PR #289, opened 2021-07-03
Fix writing C64 disks

davidgiven:
This should hopefully make writing C64 disks on 80-track drives work again.

Fixes: #288


wybren1971:
See my test at #288 
it works fine for 80 track drives. writing an .d64 image back to floppy with a 40 track drive fails because a track gets skipped each time. Fluxengine assumes a 80 track drive.

PR #291, opened 2021-07-04
Added option for 40-track drive

wybren1971:
Hello @davidgiven , i added some code to be able to use my 40 track drive to write .d64 files.
i hope i used the protoconfig correctly.

please review my code.

thanks Wybren

### DIFF
--- a/arch/c64/c64.proto
+++ b/arch/c64/c64.proto
@@ -9,5 +9,6 @@ message Commodore64EncoderProto {
 		(help) = "post-index gap before first sector header."];
 	optional double clock_compensation_factor = 2 [default=1.0,
 		(help) = "scale the output clock by this much."];
+	optional bool forty_track_drive = 3 [default=false, (help) = "is a 40 track drive used."];
 }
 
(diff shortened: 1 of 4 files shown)


wybren1971:
Is it an option to include the forty_track_drive to config.proto? like so:
```
message ConfigProto {
	optional string comment = 8;
	optional bool forty_track_drive = 9;
	
	optional InputProto input = 1;
	optional OutputProto output = 2;
	optional EncoderProto encoder = 3;
	optional DecoderProto decoder = 4;
	optional UsbProto usb = 5;

	optional RangeProto cylinders = 6;
	optional RangeProto heads = 7;
}'
``` 
In the code can than be checked if a 40 track drive is attached.
users with a forty track drive can change that parameter and everything works.

davidgiven:
That should work, but it's not actually the way I want to do it --- I want to put a final stage translator in the reader and writer which deals with the conversion. This will allow most of the client to behave as if it's connected to a 80-track drive rather than every encoder having to do the work. I'll look into this myself shortly; there's another change I want to do first. Thanks, though!

PR #292, opened 2021-07-05
Fix distorted picture

wybren1971:
modify available space (width and height) with subtraction of two times the border (up en down, and left and right)


davidgiven:
Thanks!

PR #296, opened 2021-07-07
Northstar: Handle mixed-density disks

hharte:
North Star DOS provided an upgrade path for disks formatted as
single-density to hold double-density data without reformatting.  In
this case, the four directory blocks will be single-density but other
areas of the disk are double-density.  This cannot be accurately
represented using a .nsi file, so in these cases, we pad the sector to
512-bytes, filling with spaces.


davidgiven:
That's great; thanks.

Issue #307, opened 2021-07-20
Build failed !


bugjacobs:
Hello !! Trying to build the fluxengine client software on Linux Mint 19.3 with the listed dependencies installed - come across this error after running "make" in directory !

Heeeeelp !!
EDIT: Im planning to use the client software for Greaseweazle ! (F7 v3 Thunderbolt) - not connected at time of build (?)

Greets Dag E

MKNINJA .obj/build.ninja
[61/555] PROTO arch/aeslanier/aeslanie...er/imagewriter.proto lib/usb/usb.proto
[...]

Issue #185, opened 2020-09-08
Trying to write IBM with tweaks for Commodore 1581


hpingel:
Hi,

it's me, I'm back - the guy with the interest in Commodore 1581. ;-)
I did some read tests last year:
https://github.com/davidgiven/fluxengine/issues/95

But now I want to try to write: So I currently use the latest Cypress image and a self-compiled version of Fluxengine (current master branch) where I have tweaked the ibm-preset-720 to my needs. Basically the Commodore 1581 disks have 10 sectors per track (s=10) and have the sides inverted.

Because of the additional sector I seem to need to decrease the gap sizes because otherwise I exceed the allowed number of bytes per track. I am completely unsure at the moment if I nee the IAM stuff or not and if I have to modify the IDAM / DAM bytes.

At the moment I can write a d81 image to disk and read it back again with Fluxengine. But the real 1581 doesn't want to read from that disk. I am wondering if the swapped sides could be an issue? It wasn't back then when I asked the same regarding reading a disk (https://github.com/davidgiven/fluxengine/issues/102).

My current settings for the 1581 are (gap values I don't provide at the moment because I'm constantly chaging them for testing purposes and it doesn't work anyway at the moment):

		setWriterDefaultInput(":c=80:h=2:s=10:b=512");
		trackLengthMs.setDefaultValue(200);
		sectorSize.setDefaultValue(512);
		emitIam.setDefaultValue(true);
		clockRateKhz.setDefaultValue(250);
		idamByte.setDefaultValue(0x5554);
		damByte.setDefaultValue(0x5545);
                 ...
		sectorSkew.setDefaultValue("0123456789");

hpingel:
I was just successful to read a disk image written by Fluxengine in my Commodore 1581 disk drive!!! Yeah!

I will create a pull request. :)

hpingel:
Thank you for merging my PR regarding swapSides (https://github.com/davidgiven/fluxengine/pull/214#issue-556335808). I hope that the ordinary ibm presets still do work now as I have not added the param to the stock ibm presets (yet). Will test if it defaults to false.

I will create another PR with suggestions for a preset for Commodore 1581 (and maybe also CBM-FD2000). As things a still under development i can't promise the perfect gap values yet.

As far as I can see the gap1 value will not change anything as long as emitIAM is false? Could we reflect that in the gap name? gapIAM maybe? Just to point out that there is a connection?

My current WIP preset looks like this:

[...]

PR #310, opened 2021-07-21
The 1581 format doesn't need swap_sides.

davidgiven:
eom

Fixes: #307

### DIFF
--- a/src/formats/commodore1581.textpb
+++ b/src/formats/commodore1581.textpb
@@ -34,15 +34,12 @@ encoder {
 			gap2: 22
 			gap3: 34
 			sector_skew: "0123456789"
-			swap_sides: true
 		}
 	}
 }
 
 decoder {
-	ibm {
-		swap_sides: true
-	}
+	ibm {}
 }
 
 cylinders {


hpingel:
Hi David,

I am convinced that 1581 is the only format on Fluxengine that actually needs swap_sides...

Cheers,
H.

davidgiven:
I had somebody with a third-party D81 image that wouldn't work in real hardware unless it was removed.

I wonder if there's something wrong with the D81 image reader that's getting the side byte inverted?

hpingel:
I will test it and get back to you. Maybe Sunday.

hpingel:
Hi David,

TLDR: Your current source code works for me fine with commodore1581 reading and writing.

But ATM I don't understand why you can afford to turn swap_sides off. My test tonight was:

1) Choosing a real floppy disk that contained known-good data that works fine in a real 1581 drive.
2) Reading this floppy disk into a d81 image.
3) Opening this d81 image in dirmaster (https://style64.org/dirmaster) or vice emu (my NVidia driver was auto-updated and I'm too lazy to reboot my Linux box right now, vice fails because of GL error) and seing the contents.
4) Writing the image back to a floppy disk using commodore1581 preset.
5) Reading the newly written image and comparing md5/sha1 sums for all the d81 images.

And it all worked.

hpingel:
Changed my mind yet again: swap_sides is needed!

bugjacobs:
I will have to doublecheck how my 1581 drive acts on this - I believe I did test it there in addition to reading and writing via GreaseWeazle hardware on my Linux install (and then reading the resulting d81 in Dirmaster and droiD64. (java software for diskimages viewing.) I know I used a bit of both GWzle software and the HxC software to convert a d81 to scp to try write it back to disk with GWzle before David made the suggestion to remove both swap_sides. At that time I could both read and write d81 with Fluxengine. Will get into this again as soon as I can its too late at night right now. :-)

hpingel:
The ususal workflow for me is to write a D81 image file to floppy disk via Fluxengine (while I use both Cypress hardware and Greaseweazle) and then use that floppy disk in my real Commodore 1581.

I don't use SCP files so far. So if the sides would be mixed up in a source file like an SCP file that could explain why it appears that swap_sides is hindering. But that would be the wrong conclusion.

My opinion is: We need swap_sides=true for writing and my opinion is based on the following:

1) Experience with writing a 1581 decoder for Rob Smiths project in 2019: https://github.com/hpingel/pyAccess1581  - see param swapsides in https://github.com/hpingel/pyAccess1581/blob/master/access1581/diskformats.py

2) https://github.com/Distrotech/fdutils/blob/master/src/mediaprm

3) Implementing swap_sides for Fluxengine as a Pull Request in January: https://github.com/davidgiven/fluxengine/pull/214
In that pull request is a detailed explanation why swap_sides is needed and for which drive formats it is needed.

Further reading:
https://github.com/davidgiven/fluxengine/issues/185
https://github.com/davidgiven/fluxengine/pull/218

I have been using Fluxengine since January with writing 1581 disks with swap_sides=true and using the disks successfully in my Commodore 1581 drive.

hpingel:
I'm doing some small read and write tests in my lunchbreak now: 

I can see where the confusion is coming from. In commodore1581.textpb there were two occurences of swap_sides for encoder and decoder. Just removing the one in the decoder breaks the on-the-fly-verify on disk write. So if I want to have swap_sides in the encoder I also have to add it to the decoder.

But: Now writing the disk works fine (checked in a real 1581) but reading a disk results in a wrong d81 image where the sides are swapped. So we need some logic to break up those dependencies.

For tests, I would recommend to start off with a valid 3rd party d81 image that could be found on csdb.dk or hvsc or zimmers.net. See here: http://www.zimmers.net/anonftp/pub/cbm/demodisks/drives/index.html

davidgiven:
The decoder historically didn't need swap_sides because it was autoconfiguring --- it read all the data and assembled an image based on the sector IDs on the disk. But, the verifier doesn't do that: it reads a single track, so it's important it needs the right one.

If the encoder needs swap_sides, then the decoder should too. I think something's getting physical and logical sides mixed up. I'm investigating.

davidgiven:
Turns out that last statement was wrong --- swap_sides isn't symmetrical. See #313.

hpingel:
Thank you for looking into this, David, I really enjoy using your project Fluxengine!

bugjacobs:
Whatever was done to the code on the current build seems to work fine for me too ! I just read a disk with GreaseWeazle that I previously formatted and wrote a small test file to in my 1581, and it read fine with the old build (the one with both swap sides removed) and then I wrote it back again with the current build and read it back again and it was still good ! Will test more in time !

davidgiven:
Excellent --- thanks very much for testing!

PR #332, opened 2021-10-02
Redirect dd to stdout instead of file '1'

ejona86:
Looks like a typo, that would have had mostly the intended effect.


davidgiven:
That is correct. Thanks.

PR #333, opened 2021-10-03
Add Micropolis encoder

ejona86:
I wrote this much earlier this year, but by the time I had tested it on a real machine the protobuf changes were already underway.

I spent time trying to get encodedecodetest working for micropolis, and was successful, but inserting the index at the right place really benefits from an `appendInterval()` at the end of `Fluxmap::appendBits()`. It seemed that could maybe mess something up, so I removed it from this PR. For example, the encoders just check that `cursor` didn't overflow, but there could be a lot of unused bits at the end of `bits`. It might also confuse some tooling to see an interval without any following pulse. Or it might all be fine. But I decided to get this part in before worrying about it.

CC @hharte


hharte (approved):
Looks good.  One point about the Micropolis in general is that for some operating systems like Micropolis MDOS and Vector MZOS, the metadata in the sector header needs to be preserved.  This is what the .vgi file format is for.  The straight .img files are fine for CP/M.

davidgiven:
Sorry for the delay --- been on holiday and catching up. Thanks very much for this! Only one quibble: could you add a encoderdecoder test for it, please? This will do a round trip of encoding some random data and decoding it again to make sure you get the same data. Look at the list at the bottom of mkninja.sh and the proto files in scripts.

ejona86:
@davidgiven, did you see my note in the PR description? Basically, for the encodedecodetest the encoder needs to add index marks to the output. Most of the index marks aren't much of a problem, but the last index mark is in the middle of the last sector... in the middle of user data. This isn't so bad, except that means I need to do multiple `appendBits()` calls per track and that function doesn't `appendInterval()` for trailing zeros. The fix is easy to code, but harder to determine if something might be accidentally depending on the previous behavior.

davidgiven:
Oh, oops --- no, I missed that, sorry. Will merge.

Re `appendBits()`: that's a nice catch. I'd completely forgotten about the end-of-track index mark being half-way between the sector index marks. Yes, adding appendInterval _ought_ to be the right thing to do but I agree about edge cases. Hopefully there are enough tests to catch most of these, but apparently I never actually got round to writing a Fluxmap test...

Issue #322, opened 2021-08-21
Micropolis disk captured as .scp on greaseweazle - any hope?


cube1us:
I captured a .scp image using a greaseweazle (V0.27) using this command using a Micropolis 1042 mod I (SSDD, 35 tracks, 143KB total capacity)

gw.exe read --drive 0 --rpm 300 --rate 250 --tracks c=0-34:h=0 --revs 60 disk38.scp  1>disk38.log 2>&1

I then tried to run it through fluxengine thusly and with these results - no .img file was created:

fluxengine read config.micropModI -s disk38.scp -o disk38.img
SCP tracks 0-34, heads 0-0
SCP sample resolution: 25 ns
  0.0: Error: SCP read I/O error: No error   <<<<<<<<<<<<<<

Did I miss something?  If not, any hope of this working, or is it a dead end (in which case I can certainly acquire the necessary hardware to use fluxengine directly.)

The profile I used (config.micropModI) was as follows:

decoder {
  micropolis {
  }
}
cylinders {
  start: 0
  end: 34
}
heads {
  start: 0
  end: 0
}
comment: "Micropolis Mod I MetaFloppy 143KB 5.25\" SSDD hard-sectored (ro)"
image_writer {
  filename: "micropolis.img"
  img {
  }
}
flux_source {
  drive {
    drive: 0
    hard_sector_count: 16
  }
}
flux_sink {
  drive {
    drive: 0
  }
}

davidgiven:
That looks very much like a bug at my end, judging by the error. Is it possible to attach the SCP file? One thing you could also try is using the FluxEngine client to read directly from the GreaseWeazle --- you will need to tell it which tty port it's on but it should work fine.

cube1us:
Reading hard sector using the fluxengine client via gw hardware is not supported - or so the error message it gave told me.    I seem to have deleted the Mod I (disk 38 scp), but here is a Mod II .scp (disk50 - a bit more than twice as many tracks - but gave the same error message), along with the gw capture log.  The profile used for this one was:

decoder {
  micropolis {
  }
}
cylinders {
  start: 0
  end: 76
}
heads {
  start: 0
  end: 1
}
comment: "Micropolis MetaFloppy 630kB 5.25\" DSDD hard-sectored (ro)"
image_writer {
  filename: "micropolis.img"
  img {
  }
}
flux_source {
  drive {
    drive: 0
    hard_sector_count: 16
  }
}
flux_sink {
  drive {
[...]

PR #335, opened 2021-10-09
scpfluxsource: Support more than 5 revolutions

ejona86:
This prevents uninitialized memory reads for files with many
revolutions. See #322.

### DIFF
--- a/lib/scp.h
+++ b/lib/scp.h
@@ -27,7 +27,7 @@ enum
     SCP_FLAG_FOOTER     = (1<<5)
 };
 
-struct ScpTrackStart
+struct ScpTrackHeader
 {
     char track_id[3];      // 'TRK'
     uint8_t strack;        // SCP track number
@@ -42,9 +42,8 @@ struct ScpTrackRevolution
 
 struct ScpTrack
 {
-    char track_id[3];      // 'TRK'
-    uint8_t strack;        // SCP track number
-    struct ScpTrackRevolution revolution[5];
+    ScpTrackHeader header;
+    ScpTrackRevolution revolution[5];
 };
 
 #endif
--- a/lib/scp.h
+++ b/lib/scp.h
@@ -27,17 +27,24 @@ enum
     SCP_FLAG_FOOTER     = (1<<5)
 };
 
+struct ScpTrackStart
+{
+    char track_id[3];      // 'TRK'
+    uint8_t strack;        // SCP track number
+};
+
+struct ScpTrackRevolution
+{
+    uint8_t index[4];  // time for one revolution
+    uint8_t length[4]; // number of bitcells
+    uint8_t offset[4]; // offset to bitcell data, relative to track header
+};
+
 struct ScpTrack
 {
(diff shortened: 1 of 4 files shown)


lib/scp.h
```diff
@@ -27,17 +27,24 @@ enum
     SCP_FLAG_FOOTER     = (1<<5)
 };
 
+struct ScpTrackStart
```
davidgiven, line 30: ScpTrackHeader, please.

lib/scp.h
```diff
@@ -27,17 +27,24 @@ enum
...
+struct ScpTrackRevolution
+{
+    uint8_t index[4];  // time for one revolution
+    uint8_t length[4]; // number of bitcells
+    uint8_t offset[4]; // offset to bitcell data, relative to track header
+};
+
 struct ScpTrack
 {
     char track_id[3];      // 'TRK'
```
davidgiven, line 45: Could probably be a ScpTrackHeader here.

lib/scp.h
```diff
@@ -27,17 +27,24 @@ enum
...
     uint8_t strack;        // SCP track number
-    struct
-    {
-        uint8_t index[4];  // time for one revolution
-        uint8_t length[4]; // number of bitcells
-        uint8_t offset[4]; // offset to bitcell data, relative to track header
-    }
-    revolution[5];
+    struct ScpTrackRevolution revolution[5];
```
davidgiven, line 47: No need for `struct` --- in C++ typedefs and struct names share a namespace.

davidgiven (changes requested):

davidgiven:
Sorry for the delay --- been on holiday! Thanks very much for this; I've requested some nitpicky changes and then it should be good to go.

lib/scp.h
davidgiven: Although, taking another look at it, doesn't your change mean that ScpTrack is never used any more? If so, remove it, please?

ourIThome:
You might want to be careful about increasing the revolution count on scp files.  I believe Jim Drew's spec lists 5 revolutions.  Also it might break Hxc if it is also hard coded to 5 revolutions?  I have emailed Jim to ask.

ourIThome:
Just seen the other thread where your already been talking with Jim about scp revolutions.

ejona86:
Note that this change only updates the scp source to support more revolutions. The scp sink still only outputs up to 5 revolutions.

lib/scp.h
ejona86: ScpTrack is still used by scpfluxsink. I swapped to ScpTrackHeader here and updated scpfluxsink.

I didn't earlier either to avoid (minor) changes to the sink, or to avoid alignment concerns (yes, it happens to work fine in this case because ScpTrackHeader is 4 bytes, but that stuff is subtle). Don't recall which.

ejona86 (commented):
Completed the tweaks. Please take another look.

davidgiven:
LGTM. Merging. Thanks!

PR #347, opened 2021-11-26
Add Japanese PC 1.2MB format.

tdaede:
This format is used by the PC-98 and Sharp X68000 series.

3.5" disks are spun at 360rpm to match the clock rate of 5.25".

### DIFF
--- a/mkninja.sh
+++ b/mkninja.sh
@@ -416,6 +416,7 @@ FORMATS="\
     hplif770 \
     ibm \
     ibm1200_525 \
+    ibm1232 \
     ibm1440 \
     ibm180_525 \
     ibm360_525 \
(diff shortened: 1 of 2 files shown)


davidgiven:
Fantastic! I actually have a borrowed X68000 waiting for a new power supply, so I could do this... (Also one of my USB floppy drives supports this format on 3.5", complete with changing the motor spin speed.)

One note: could you add an encodedecode test for it, please? It should just be a matter of adding it to the list in mkninja.sh.

tdaede:
Added.

Yeah most Teac USB floppy drives (including the IBM Thinkpad ones) are trimode, as are most floppy drives on later PC-9821 series. There is a pin on trimode floppy drives to select speed (pin 1 on a 34 pin cable iirc).

You don't actually need to spin the drive faster though, scaling the bit clock also works. (I haven't tried this with fluxengine, but it works with greaseweazle)

davidgiven:
Great --- thanks!

I wonder if it's worth supporting that pin for marginally faster reads and writes on drives which support it...

tdaede:
It does also push the needed min/max flux periods into more typical values, *maybe* making for better reads, but I haven't actually observed this yet.

PR #348, opened 2021-11-26
Add DIM format reader.

tdaede:
This format is common for Japanese PC floppy images.


davidgiven:
Thanks! As always, one nit-pick: could you add the paragraph in 'Image sources and destinations' in `doc/using.md` describing this, please?

davidgiven:
Yup, that looks fine. Thanks!

PR #350, opened 2021-11-29
Fix inverted density select in Greaseweazle driver.

tdaede:
Pin 2 should be set high when high density is selected.

### DIFF
--- a/lib/usb/greaseweazleusb.cc
+++ b/lib/usb/greaseweazleusb.cc
@@ -515,7 +515,7 @@ public:
     {
         do_command({ CMD_SELECT, 3, (uint8_t)drive });
         do_command({ CMD_MOTOR, 4, (uint8_t)drive, 1 });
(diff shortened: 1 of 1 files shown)


davidgiven:
Are the values provided to the GreaseWeazle voltage levels (1 == deasserted, 0 == asserted) or logic levels (1 == asserted, 0 == deasserted)?

tdaede:
Seems to be voltage levels (the drivers are open collector, though). At least on all of my 5.25" drives, I want a high voltage level (logical 0) on pin 2 for HD media.

tdaede:
FWIW I discovered this because my drives were giving either very poor or unusable reads, so I can at least say that this polarity solves my problems. It also seems to match the datasheets that I've found for 5.25" drives.

davidgiven:
I think my drives autodetect this. I've never observed any change in behaviour from setting or unsetting the flag. This might explain oddities that some others have observed --- thanks.

tdaede:
Yeah, it's mostly important for 5.25" drives, which have no method of autodetection.

PR #351, opened 2021-11-30
Add alias for XDF image files.

tdaede:
This is a common extension for Sharp X68000 bare disk images.


davidgiven:
LGTM --- thanks!

PR #357, opened 2021-12-01
Add minimal D88 image format reader.

tdaede:
This implements a subset of D88, only supporting the first floppy
in an image. It only supports images with equal sector sizes,
equal numbers of sectors per track, and no bad sectors.


davidgiven:
Looks good. Thanks (and for the other PRs).

tdaede:
Rebased.

davidgiven:
Looks good.

PR #358, opened 2021-12-01
Refactor to allow image readers to specify the encoder/decoder formats, if desired.

### LINKED ISSUE #352: Image-defined geometry
I'm working on adding a d88 disk image reader, which is basically the Japanese equivalent of IMD format - each sector can specify its size and encoding independently. This is important as PC-88 disks frequently have different sector sizes on their first and subsequent track.

Currently, fluxengine requires you to specify a write profile which has explicit geometry - e.g. the ibm1232 profile. I was wondering if it would be beneficial to add an "auto" profile (or repurpose the "ibm" profile that's currently ro) for writing these sort of images, and source all of the profile parameters from the image itself.

Less importantly, such a mode might be handy for e.g. dim files that basically imply ibm1232 format, fluxengine could automatically pick a suitable profile.

davidgiven:
Fixes #352

### DIFF
--- a/lib/imagereader/imagereader.h
+++ b/lib/imagereader/imagereader.h
@@ -28,15 +28,10 @@ public:
     static std::unique_ptr<ImageReader> createFdiImageReader(const ImageReaderProto& config);
 
 public:
-	const Image& readImage();
-	void preloadImage();
-
-protected:
-	virtual std::unique_ptr<Image> readImageImpl() = 0;
+	virtual std::unique_ptr<Image> readImage() = 0;
 
 protected:
 	const ImageReaderProto& _config;
-	std::unique_ptr<Image> _image;
 };
 
 #endif
--- a/lib/imagereader/imagereader.h
+++ b/lib/imagereader/imagereader.h
@@ -1,9 +1,10 @@
 #ifndef IMAGEREADER_H
 #define IMAGEREADER_H
 
+#include "image.h"
+
 class ImageSpec;
 class ImageReaderProto;
-class Image;
 
 class ImageReader
 {
@@ -27,10 +28,15 @@ public:
     static std::unique_ptr<ImageReader> createFdiImageReader(const ImageReaderProto& config);
 
 public:
-	virtual Image readImage() = 0;
+	const Image& readImage();
+	void preloadImage();
+
+protected:
+	virtual std::unique_ptr<Image> readImageImpl() = 0;
 
 protected:
 	const ImageReaderProto& _config;
+	std::unique_ptr<Image> _image;
 };
 
 #endif
--- a/lib/imagereader/imdimagereader.cc
+++ b/lib/imagereader/imdimagereader.cc
@@ -90,7 +90,7 @@ public:
         ImageReader(config)
     {}
 
-    std::unique_ptr<Image> readImageImpl()
+    std::unique_ptr<Image> readImage()
     /*
     IMAGE FILE FORMAT
     The overall layout of an ImageDisk .IMD image file is:
--- a/lib/imagereader/imdimagereader.cc
+++ b/lib/imagereader/imdimagereader.cc
@@ -89,7 +89,7 @@ public:
 		ImageReader(config)
 	{}
 
-	Image readImage()
+	std::unique_ptr<Image> readImageImpl()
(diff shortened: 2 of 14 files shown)


lib/imagereader/imagereader.h
```diff
@@ -27,10 +28,15 @@ class ImageReader
     static std::unique_ptr<ImageReader> createFdiImageReader(const ImageReaderProto& config);
 
 public:
-	virtual Image readImage() = 0;
+	const Image& readImage();
+	void preloadImage();
+
+protected:
+	virtual std::unique_ptr<Image> readImageImpl() = 0;
```
tdaede, line 35: Is the std::unique_ptr just there to reduce copies or does something else require it?

src/fe-write.cc
```diff
@@ -62,6 +62,8 @@ int mainWrite(int argc, const char* argv[])
     flags.parseFlagsWithConfigFiles(argc, argv, formats);
 
 	std::unique_ptr<ImageReader> reader(ImageReader::create(config.image_reader()));
+	reader->preloadImage();
```
tdaede, line 65: Is there any reason you can't just load the Image here and then  pass that to writeDiskCommand rather than an ImageReader?

lib/imagereader/imagereader.h
davidgiven: It's just to remove copies.

src/fe-write.cc
davidgiven: That's a very good point --- passing the image to writeDiskCommand makes a lot more sense as the internal interface works with Image anyway, and this gets rid of some reader state.

tdaede:
Looks good. If you just want to land it without the IMD part I can try writing a version on one of the simple readers like dim or fdi that just have a list of 3 possible track formats.

(I'm a bit tempted to actually do that by reading the compiled in format definitions)

lib/imagereader/imdimagereader.cc
```diff
@@ -85,204 +86,196 @@ static unsigned getSectorSize(uint8_t flags)
...
+                    mfm ? "MFM" : "FM",
+                    Modulation_Speed, header.numSectors, sectorSize, sector_skew, (header.track+1) * trackSize / 1024);
+
+        if (config.encoder().format_case() == EncoderProto::FormatCase::FORMAT_NOT_SET)
+        {
+            std::cout << "IMD: setting encoder format automatically\n";
+
+            auto ibm = config.mutable_encoder()->mutable_ibm();
+            auto trackdata = ibm->add_trackdata();
+            trackdata->set_track_length_ms(200);
```
tdaede, line 252: What order do the command line options currently get applied in? e.g. is it possible to override these settings?

lib/imagereader/imdimagereader.cc
```diff
@@ -85,204 +86,196 @@ static unsigned getSectorSize(uint8_t flags)
...
-					header.track, header.Head + 1,
-					mfm ? "MFM" : "FM",
-					Modulation_Speed, header.numSectors, sectorSize, sector_skew, (header.track+1) * trackSize / 1024);
+        std::cout << fmt::format("IMD: {} tracks, {} heads; {}; {} kbps; {} sectors; sectorsize {};\n"
+                                 "     sectormap {}; {} kB total\n",
+                    header.track, header.Head + 1,
+                    mfm ? "MFM" : "FM",
+                    Modulation_Speed, header.numSectors, sectorSize, sector_skew, (header.track+1) * trackSize / 1024);
+
+        if (config.encoder().format_case() == EncoderProto::FormatCase::FORMAT_NOT_SET)
```
tdaede, line 246: Convention wise I think it is OK for the specified profile to override the format's internal profile for logical image formats like img/fdi/dim, but I think for formats like imd and d88 that have explicit sector encodings, that should always override the profile.

lib/imagereader/imdimagereader.cc
davidgiven: Yup. The way it's set up, the automatic settings are only applied if no encoder/decode is provided in the profile. So `write ibm1440 -i something.imd` will always use the `ibm1440` profile, and `write -i something.imd` will use the IMD profile.

PR #359, opened 2021-12-02
Add Make target to build compdb.

tdaede:
Add Make target to build compdb.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,3 +64,5 @@ clean:
 	@mkdir -p $(OBJDIR)
 	@sh $< > $@
 
+compdb:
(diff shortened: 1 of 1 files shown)


tdaede:
Specifically for clangd, which I use with emacs but is the same thing used in vscode.

I can make it rebuild every time. The "canonical" location for compile_commands.json is in the root of the project directory or build/, so if it's emitted into .obj the user will need to make a symlink from the root directory if they want to use it.

davidgiven:
Belated thanks.

tdaede:
No problem.

If you have time, I'd like a review of #370, as I have more patches that depend on that one.

PR #361, opened 2021-12-03
Automatically configure format for DIM when no format is specified.

tdaede:


lib/imagereader/imagereader.h
```diff
@@ -28,10 +29,13 @@ class ImageReader
     static std::unique_ptr<ImageReader> createFdiImageReader(const ImageReaderProto& config);
 
 public:
-	virtual std::unique_ptr<Image> readImage() = 0;
+	virtual std::unique_ptr<Image> readImage(EncoderProto* encoder);
```
davidgiven, line 32: I'm in two minds about passing in the encoder like this. On the plus side, mutating global state is kinda bad. On the minus side, there may be other things that need changing as well --- the most obvious being the decoder, but you may also want to set `heads` and `cylinders` to a sensible default if not already set. I think I'd rather go with mutating global state, TBH, as it's a less intrusive change.

config is a global (available from protos.h) so you don't need to pass this in.

davidgiven (changes requested):

lib/imagereader/imagereader.h
tdaede: Updated.

davidgiven (approved):

davidgiven:
All LGTM --- thanks!

PR #364, opened 2021-12-03
Improve SCP handling.

### LINKED ISSUE #362: SCP files created with Greaseweazle's encoder have a bunch of zero-length revolutions
This confuses tools like HxC and Greaseweazle:

https://github.com/keirf/Greaseweazle/issues/131

I'm not sure why SCP writes 5 revs when the default # of revs to read is 1.2.

davidgiven:
Emit the right number of revolutions in the output SCP file. Also add some more SCP testing.

Fixes #362


lib/fluxsink/scpfluxsink.cc
```diff
@@ -138,6 +137,7 @@ class ScpFluxSink : public FluxSink
 			}
 		}
 
+		_fileheader.revolutions = revolution - 1;
```
tdaede, line 140: I think this is off by 1

davidgiven: Yes, you're right.

davidgiven: It's worse than that, there's no `break` inside the loop, so it always writes five revolutions...

PR #365, opened 2021-12-03
Write only 1 revolution with SCP.

tdaede:
A better fix would be to look at the Fluxmap length, but the
Fluxmap lacks a rotation period so that's a more involved fix.

### DIFF
--- a/lib/fluxsink/scpfluxsink.cc
+++ b/lib/fluxsink/scpfluxsink.cc
@@ -47,7 +47,7 @@ public:
 		_fileheader.file_id[2] = 'P';
 		_fileheader.version = 0x18; /* Version 1.8 of the spec */
 		_fileheader.type = _config.type_byte();
-		_fileheader.revolutions = 5;
+		_fileheader.revolutions = 1;
 		_fileheader.start_track = strackno(config.cylinders().start(), config.heads().start());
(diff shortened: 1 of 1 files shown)


davidgiven:
I've actually got a fix for this in #364.

tdaede:
That improves things, however HxC software still struggles to decode revolutions > 1 and guesses RPM wrong. Not entirely sure if that's HxC's fault or not.

davidgiven:
In fact, I think using one revolution is better. The SCP format doesn't allow tracks to have different number of revolutions, which is a big problem. Limiting it to one is kinda bad but I think it's going to be less problematic. So I'll back out my change and use this instead.

davidgiven:
The changes are in #367.

tdaede:
In what situation would you want tracks to have different numbers of revolutions?

davidgiven:
If there are transient errors during a track read, the read will be retried, leading to more flux data. Although, it's not really more _revolutions_, instead being a completely different read, which I don't think scp can represent...

Also currently the `.flux` writer will discard all but the last read, and the `.scp` writer will write multiple track structures with the index pointing at the last one. That all need fixing.

tdaede:
Ah I see, yeah the revolutions is only per disk. The SCP writer will need the same logic as the .flux writer.

PR #367, opened 2021-12-03
Writing more than one SCP revolution is problematic, so don't.

davidgiven:


tdaede:
This is at least more functional for using scp files created with the encoders.

(I think for archival flux backups using the SCP format, storing multiple revolutions can still be useful, especially if the read is not index pulse aligned. In particular the SCP writer needs to read the fluxmap to determine the number of revolutions, and then truncate to an integer number)

tdaede:
Thinking about this a bit more, considering the default of not aligning to index pulse, I think this needs a really fat warning that the SCP writer should not be used for archival at the moment. (though in its current state before the patch, the warning is probably also justified).

Are you working on a more permanent fix? If not I can look at it.

(I have no particular love for the SCP format, but it's the gw tools' "native" format and an easy way to debug with hxc)

tdaede:
I think this is no longer needed with #380.

davidgiven:
Cool. Was just wondering that.

PR #369, opened 2021-12-04
Use per-track encoding specified in D88 files.

tdaede:
This supports varying sector sizes, counts, and modulation on
a per-track basis. It also supports writing sectors in the
original order.

### DIFF
--- a/doc/using.md
+++ b/doc/using.md
@@ -276,8 +276,8 @@ FluxEngine also supports a number of file system image formats. When using the
   FluxEngine is currently limited to reading only the first floppy image in a
   D88 file.
   
-  In addition, it only supports a limited subset of D88 features and will
-  reject files not in that subset.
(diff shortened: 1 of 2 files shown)


src/fe-write.cc
```diff
@@ -68,7 +68,8 @@ int mainWrite(int argc, const char* argv[])
 	std::unique_ptr<FluxSink> fluxSink(FluxSink::create(config.flux_sink()));
 
 	std::unique_ptr<AbstractDecoder> decoder;
-	if (config.has_decoder())
+	//if (config.has_decoder())
```
davidgiven, line 71: Stray debugging code.

davidgiven (changes requested):

davidgiven (approved):

davidgiven:
Thanks!

PR #370, opened 2021-12-04
Add option to rescale flux sink pulses. 

tdaede:
Note that this does not affect the decoder used for verify,
but at least for tame rescaling the decoders are still able to
lock.

This is a precursor to automatic scaling via the disk speed
measurement.

### DIFF
--- a/doc/using.md
+++ b/doc/using.md
@@ -338,6 +338,14 @@ behaviour.
 	has no effect on the _drive_, so it doesn't help with flippy disks, but is
 	useful for using very old drives with FluxEngine itself. If you use this
 	option, then any index marks in the sampled flux are, of course, garbage.
+  
+  - `--flux_sink.rescale=X`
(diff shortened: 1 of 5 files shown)


tdaede:
rebased

tdaede:
Any thoughts on this? This is the last thing really needed for fully functional PC-98 support.

tdaede:
(well, that and the equivalent option for flux sources, which I'll write after this)

davidgiven:
I'm sorry for the long delay on this...

I'll merge this, but I actually think this use case is better served by adjusting the clock in the encoder. It ought to be possible to tell the encoder the target rotational rate and clock speed, and have it figure out the required clock for the current drive rotational rate from that --- you can see some thoughts towards that in the Victor 9000 encoder. That way there's never any need to actually rescale a generated fluxmap.

PR #371, opened 2021-12-04
Set media density automatically for D88 format.

tdaede:


davidgiven:
I was wondering whether letting the encoder set a drive property was a good idea, but I think on the balance, yes.

tdaede:
I don't really love this method of doing it, I'd rather the internal Image carry a property over to the encoder, and eventually the writer. I also need this for double-stepping.

Issue #368, opened 2021-12-04
FM pre-data filler bits are wrong


tdaede:
This doesn't actually have an effect on reading, but comparing a floppy written natively by N88-BASIC on my PC-9821:

![Screenshot from 2021-12-03 16-47-08](https://user-images.githubusercontent.com/1308314/144690131-aad5c836-39f2-49de-9b26-c1c503c271ea.png)

The version produced by Fluxengine's FM code fills with half the clock rate, and lacks lead-in clocks:

![Screenshot from 2021-12-03 16-48-02](https://user-images.githubusercontent.com/1308314/144690156-78c2ae97-a546-4a08-860c-2123025613bd.png)

davidgiven:
What are the graphs showing --- intervals on the Y axis? That would make the fill 0101010... on the FluxEngine, and 11111... on the PC-9821. That makes sense, as I'm using 0x00 encoded to FM as the fill byte, as described in https://www.hermannseib.com/documents/floppy.pdf page 9. It sounds like you PC-9821 is using 0xff instead. This is allowed but it makes sense to have it configurable in the profile.

What do you mean by lead-in clocks?

tdaede:
Yeah exactly. "Wrong" might be a bit of a strong word here, as this is up to the floppy chip, though I would think that the 1111 filler byte pattern would be easier for a PLL to lock onto.

The lead-in clocks are the 3 bytes of 0x00 that the PC-9821's controller writes before the sync pattern. Fluxengine doesn't write them (though if it did, they would be invisible because they are the same as the fill pattern).

tdaede:
Also the area I'm showing is gap2, which is defined on page 10 of the document you linked as having 11 bytes of 0xff followed by 6 of 0x00.

davidgiven:
Right, I definitely just saw the code that writes those six bytes, so I think they're there.

The fix should be in #374, but if you have an FM profile then this will need changing as the default value is only suitable for MFM. (We don't have any FM profiles currently...)

tdaede:
Ah yeah you're right, the 6 bytes are there.

PR #374, opened 2021-12-05
Make the gap fill byte configurable.

davidgiven:
The gap fill byte (the space between records) can now be configured in the profile.

The default value is for MFM, so when we get some FM profiles they will need to override this.

Fixes: #368


tdaede:
Looks good to me, though I think just having the filler byte be keyed on use_fm would probably be good enough for most purposes.

davidgiven:
Cool, thanks!

PR #376, opened 2021-12-06
Fix SCP writer to always write complete revolutions.

tdaede:
There were a couple of problems that needed to be fixed:

- Fluxmaps can be produced with zero index pulses by encoders.
This is somewhat ambiguous but it's assumed that this means there
is exactly one revolution.

- The SCP unindexed mode wasn't properly supported, as it should
create equal-length revolutions not aligned to indexes. However,
fluxmaps don't contain any information on rotation speed. Therefore,
drop the ability to create unindexed SCPs and slightly change the
meaning of the option to not include pre-index data.

- Real flux reads can potentially produce duplicate index pulses.
Avoid creating zero-length revs.

- Fix off-by-one error for rev count.

### DIFF
--- a/lib/fluxsink/fluxsink.proto
+++ b/lib/fluxsink/fluxsink.proto
@@ -20,7 +20,7 @@ message VcdFluxSinkProto {
 
 message ScpFluxSinkProto {
 	optional string filename = 2       [default = "flux.scp", (help) = ".scp file to write to"];
-	optional bool align_with_index = 3 [default = false, (help) = "align data to track boundaries"];
+	optional bool align_with_index = 3 [default = false, (help) = "discard data before the first index pulse"];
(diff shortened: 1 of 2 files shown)


davidgiven:
You'll probably be unthrilled to hear that I recently added a F_DESYNC event to the bytecode stream (which is represented as a zero, which used to be EOF) to represent an arbitrary gap. This is used for combining flux from multiple reads, as maintaining multiple flux streams made things too complicated.

I think you'll need to discard any revolutions containing any of these so that you're left with known complete revolutions. This probably wants something like a `FluxMap::trim()` which does the work, returning a new fluxmap, as it'd be useful elsewhere.

tdaede:
I think that's actually broken for all the flux sinks right now, as I don't think any of them really support multiple flux streams (A2R does, for when I add it). I can open a separate PR to edit them all to only use the last segment in the stream.

davidgiven:
Merged; thanks!

PR #377, opened 2021-12-06
Set FM gap fill byte in D88.

tdaede:


lib/imagereader/d88imagereader.cc
```diff
@@ -127,7 +127,7 @@ class D88ImageReader : public ImageReader
                     trackdata->set_sector_size(sectorSize);
                     trackdata->set_use_fm(fm);
                     if (fm) {
-                        //trackdata->set_clock_rate_khz(250*300/360);
+                        trackdata->set_gap_fill_byte(0xff);
```
davidgiven, line 130: Are you sure? It's a raw 16-bit value. Do you want 0xffff instead?

tdaede: Yes.

davidgiven:
Merged --- thanks!

PR #378, opened 2021-12-06
Add ibm_n88basic format.

tdaede:
This isn't really intended to be used directly (bare images are
normally .d88 etc) but is useful for testing the FM encoder.

### DIFF
--- a/mkninja.sh
+++ b/mkninja.sh
@@ -440,6 +440,7 @@ FORMATS="\
     mac800 \
     micropolis \
     mx \
+    n88basic \
     northstar175 \
     northstar350 \
     northstar87 \
@@ -541,6 +542,7 @@ encodedecodetest ibm180_525
 encodedecodetest ibm360_525
 encodedecodetest ibm720
 encodedecodetest ibm720_525
+encodedecodetest n88basic
 encodedecodetest tids990
 encodedecodetest commodore1581
 encodedecodetest commodore1541 scripts/commodore1541_test.textpb
--- a/src/formats/n88basic.textpb
+++ b/src/formats/n88basic.textpb
@@ -1,0 +1,134 @@
+comment: 'N88-BASIC 5.25"/3.5" 77-track 26-sector DSHD'
+
+flux_sink {
+	drive {
+		high_density: true
+	}
+}
+
+flux_source {
+	drive {
+		high_density: true
+	}
+}
+
+image_reader {
+	filename: "n88basic.img"
+	img {
+		tracks: 77
+		sides: 2
+        trackdata {
+			sector_size: 256
+			sector_range {
+				start_sector: 1
+				sector_count: 26
+			}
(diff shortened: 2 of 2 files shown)


src/formats/ibm_n88basic.textpb
```diff
@@ -0,0 +1,117 @@
...
+            track: 0
+            up_to_track: 0
+			sector_size: 128
+			sector_range {
+				start_sector: 1
+				sector_count: 26
+			}
+		}
+        trackdata {
+            track: 0
```
davidgiven, line 30: This range overlaps the previous trackdata clause, and matching clauses are merged top to bottom, so the `sector_size: 256` will override the 128 in the clause above.

mkninja.sh
```diff
@@ -541,6 +542,7 @@ encodedecodetest ibm180_525
 encodedecodetest ibm360_525
 encodedecodetest ibm720
 encodedecodetest ibm720_525
+encodedecodetest ibm_n88basic
```
davidgiven, line 545: I think just calling this `n88basic` will work, as this has nothing to do with standard IBM formats. This was on the PC-8800, right? Was it used by anything other than BASIC? If so, calling it `pc8800` might be better.

tdaede: I can do that too, it's still a standard IBM FDC with normal MFM/FM on it which is why I left the IBM prefix.

It's used by games too but mostly those written in BASIC. It's also used by the PC-98's implementation of N88-Basic. The PC-88 supports pretty arbitrary sector layouts, so I don't think there is any one format I'd call pc8800.

src/formats/ibm_n88basic.textpb
tdaede: Indeed, reordered the other block as well and added an imagewriter definition.

mkninja.sh
davidgiven: We have other normal IBM formats without the prefix --- Acorn ADFS and DFS, for example. The `ibm` ones are just the ones generic enough they're used by more than one platform. This one's hardware specific so I think `n88basic` is most appropriate.

src/formats/ibm_n88basic.textpb
davidgiven: Yup, LGTM.

BTW, do you have actual hardware and have verified this works? In which case this could very plausibly get a dinosaur in the big table in the README...

mkninja.sh
tdaede: It's been updated to remove the prefix.

src/formats/ibm_n88basic.textpb
tdaede: I have real hardware and confirmed that .d88 files read and written with this layout work, however I have not tested this .textpb file specifically.

davidgiven:
Merged --- thanks!

PR #380, opened 2021-12-06
SCP: Only write the last continuous fluxmap.

tdaede:


davidgiven:
Yup, that'll work. I was also thinking of a `normalise()` method which would return a fluxmap containing only complete revolutions (i.e. bounded by index marks with no desync between them) but this will do for now.

tdaede:
In SCP's (and A2R's) case we don't actually want to normalize as it allows data before and after revolutions (to store e.g. fractional revolutions). It would make sense if attempting to write the flux image to a disk, though.

PR #381, opened 2021-12-07
Fix North Star after decoder update

hharte:
Similar to Micropolis, the North Star decoder was broken by the PLL update.

Fix it and also fix the image size output in the informational message in nsiimagewriter.

Here are [example North Star flux files](https://www.dropbox.com/s/eks28s8dwsal04a/northstar_flux.zip?dl=0)

There is one each of 87.5K SSSD, 175K SSDD, 350K DSDD.

Thanks,
Howard


davidgiven:
Yes, that's wrong --- thanks for the fix.

PR #382, opened 2021-12-07
Fix a small typo in proto.cc

darkstar:
This is most certainly a typo.

Also it could be nice to have the long-forms of "y" and "n" available as possible options as well


davidgiven:
Er, oops?

Thanks!

Issue #393, opened 2021-12-17
GreaseWeazle does not work on COM port > 9 on Windows


hharte:
While experimenting with FluxEngine support for GreaseWeazle on Windows 10, I found that it did not work properly when the system assigned a COM port number > 9 to the GreaseWeazle.

davidgiven:
Are you using the up-to-date client? I ask because only a few days ago I committed a major rework to the way serial ports are handled.

davidgiven:
...aaand now I see your PR!

davidgiven:
Closing as this got fixed in the PR.

PR #394, opened 2021-12-17
Windows: Support serial ports > COM9

hharte:
Per #393 FluxEngine can not open the serial port when COM port number is greater than COM9.

Windows requires the use of the DOS UNC path name for COM ports greater than COM9.

Fix: Prefix the port name with \\\\.\\


davidgiven:
Windows is awesome. Thanks very much for this.

PR #397, opened 2021-12-25
reader: Fix conflicting sector error message.

hharte:
The conflicting sector error message incorrectly displays the track instead
of the sector in conflict.  Fix by removing unused track and head arguments,
displaying only the sector in conflict.


davidgiven:
Thanks!

PR #398, opened 2021-12-29
Micropolis raw 275 byte sectors (VGI)

ejona86:
This adds support for the VGI format that is commonly used for Micropolis disks. It is essentially just a dump of the full sector data and metadata. I then add convenience formats for the various Micropolis variations. Data-only IMG is still supported, but isn't geared toward the same convenience.

VGI is an emulator-centric format, so sectors are technically stored in physical order. However, most systems don't have interleave/skew. I have support for this prepared, which uses the sector/index pulses to determine sector numbers. But the code is a bit complex and is technically distinct from this PR. It also will need to be enhanced to observe desync opcodes. It'll be a followup.

CC @hharte


hharte:
Thanks, @ejona86, this looks good.  One suggestion: you may want to add requiredSectors() to allow FluxEngine to retry until all 16 sectors are read.

Regarding hard sector interleave, I have not seen this on any of the Micropolis disks I’ve imaged so far.  These include disks for MDOS, MZOS, CP/M, and OASIS.

ejona86:
> you may want to add requiredSectors() to allow FluxEngine to retry until all 16 sectors are read.

Yeah, I saw that, but it looked like it could be done as a separate PR just as easily. I was optimizing for sending this out sooner as I have limited time at the moment due to family.

> skew is more of a software thing that can be achieved even with a 1:1 physical to logical sector mapping.

Ah, okay. That makes sense. That is convenient to allow checking the soft and hard sector ids match (at least during development), since we know it is easy to choose the wrong sync byte.

hharte:
Sounds good.  I look forward to seeing this PR merged.

davidgiven:
I'll admit to not being particularly keen on this particular approach to producing raw FM/MFM data. It's really abusing the decoder framework to bypass the actual decode. All the information needed to get the half-decoded data is already there already, including things like sector ordering.

I think what I'd rather do is to have the output Image (which is passed to the ImageWriter) contain the Records for each sector. Then the ImageWriter can decide whether to use the decoded data or the half-decoded data. That way no changes need to happen in the decoder itself. The downside is that you'll only get data for sectors which correctly decode, so you can't use that route to get a complete MFM or FM or GCR stream for the entire track, but I think that's a different problem.

Let me take a look and get back to you.

davidgiven:
Er, whoops! I misunderstood what the VGI format was --- I thought it was an MFM dump, where it's actually not. That does actually mean that your solution here is the best one. (Although I think I still want the ability to do undecoded dumps too...)

There's a couple of nitpicky things I'd like in terms of naming the profiles, but other than that I think it's good to go.

src/formats/micropolis.textpb
```diff
@@ -1,4 +1,4 @@
-comment: 'Micropolis MetaFloppy 630kB 5.25" DSDD hard-sectored'
+comment: 'Micropolis MetaFloppy 630kB 5.25" DSDD hard-sectored IMG data-only'
```
davidgiven, line 1: For consistency with the other profiles, I'd prefer to have this be the normal profile (i.e. the VGI profiles are labelled as such, rather than the user-data-only profile being specially labelled).

davidgiven: Actually, thinking about this, there's a cleaner way. You can actually specify multiple profiles on the command line. So if by default, the profiles are configured to read user-data only to .img files, it'd then be possible to make an additional `vgi` profile which changes the filename and sector size to 275. You'd use it like:

`fluxengine read micropolis143 vgi -d something.vgi`

That means the same set of profiles can be used for normal `.img` files and the special `.vgi` files by just adding the modifier. Plus, eventually you'll want a special `.vgi` writer (in order to get the sectors in the right order), and this would be able to patch the configuration to add the `vgi` modifier automatically if you try to read from or write to a VGI file. This should simplify things a lot. WDYT?

hharte:
Hello David,

Your suggestion above sounds good in general.  I like the idea of having a separate vgi profile that can be used in combination with the micropolis* variants.

I'm not convinced we will need a separate .vgi writer.  In all the Micropolis disks I've examined, the physical to logical sector interleave is 1:1.  CP/M does have a "skew" but this is done at a software level inside the CP/M BIOS, and the sector IDs written in the sector headers match the hard sector ID.

src/formats/micropolis.textpb
ejona86: That works reasonably. I don't know if we'll ever have a VGI writer, but even without this works well.

davidgiven:
Sorry, this took way too long, but this looks mostly LGTM to me. There are a few things I want to change --- not in your code, but in the general way in which modifier profiles are handled --- so I'll merge this manually.

ejona86:
@davidgiven, sounds fine. Thanks for the review. It was speedy enough for me. I was delayed by a month, so a week is NBD.

PR #415, opened 2022-01-16
add generic gw-compatible support

ladyada:
hiya! im making a generic arduino-compatible floppy disk read/write library and one of the modes is 'greaseweazle compatibility' - this PR makes it so that fluxengine can successfully detect USB CDC serial ports when there is no exact VID/PID match, fixes a small packing bug in readflux, and also makes generic CDC windows 'activate' by enabling the DTR line. i also added zip upload on PR so folks can test without a release/build.

1) update CI to add zip on commit for quick testing
2) fix gw readflux command (should be 8 bytes long but was packed as 10)
3) add support for non-gw VID/PID (generic serial port that is gw compatible)
4) manually set/clear DTR on serial port devices - this seems to be essential for tinyusb/arduino CDC

not tested with a gw but ought not to have broken anything!

```
     Tracks -> 1         2         3
H.SS 0123456789012345678901234567890123456789
0. 1 ........................................
0. 2 ........................................
0. 3 ........................................
0. 4 ........................................
0. 5 ........................................
0. 6 ........................................
0. 7 ........................................
0. 8 ........................................
0. 9 ........................................
1. 1 ........................................
1. 2 ........................................
1. 3 ........................................
1. 4 ........................................
1. 5 ........................................
1. 6 ........................................
1. 7 ........................................
1. 8 ........................................
1. 9 ........................................
Good sectors: 720/720 (100%)
Missing sectors: 0/720 (0%)
Bad sectors: 0/720 (0%)
IMG: wrote 40 tracks, 2 sides, 360 kB total
```

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -53,3 +53,12 @@ jobs:
       run: |
         make
 
+    - name: zip
+      run: |
+        zip -9 fluxengine.zip fluxengine.exe brother120tool.exe brother240tool.exe FluxEngine.cydsn/CortexM3/ARM_GCC_541/Release/FluxEngine.hex
+
+    - name: Upload build artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: ${{ github.event.repository.name }}.${{ github.sha }}
+        path: fluxengine.zip
--- a/lib/usb/greaseweazleusb.cc
+++ b/lib/usb/greaseweazleusb.cc
@@ -131,7 +131,7 @@ public:
                     .write_8(CMD_READ_FLUX)
                     .write_8(cmd.size())
                     .write_le32(0) //ticks default value (guessed)
-                    .write_le32(2);//guessed
+                    .write_le16(2);//revolutions
                 do_command(cmd);
             }
         }
--- a/lib/usb/serial.cc
+++ b/lib/usb/serial.cc
@@ -45,8 +45,16 @@
 			COMMTIMEOUTS commtimeouts = {0};
 			commtimeouts.ReadIntervalTimeout = 100;
 			SetCommTimeouts(_handle, &commtimeouts);
+            
+			if (!EscapeCommFunction(_handle, CLRDTR))
+			  Error() << fmt::format("Couldn't clear DTR: {}",
+                                     get_last_error_string());
+			Sleep(200);
+			if (!EscapeCommFunction(_handle, SETDTR))
+			  Error() << fmt::format("Couldn't set DTR: {}",
+                                     get_last_error_string());
 
-			PurgeComm(_handle, PURGE_RXABORT|PURGE_RXCLEAR|PURGE_TXABORT|PURGE_TXCLEAR);
+            PurgeComm(_handle, PURGE_RXABORT|PURGE_RXCLEAR|PURGE_TXABORT|PURGE_TXCLEAR);
 		}
 
 		~SerialPortImpl() override
--- a/lib/usb/usb.cc
+++ b/lib/usb/usb.cc
@@ -69,7 +69,9 @@ USB* get_usb_impl()
 			return createGreaseWeazleUsb(candidate->serialPort, config.usb().greaseweazle());
 
 		default:
-			Error() << "internal";
+			std::cerr << fmt::format("Using Unknown GW Compatible {} on {}\n", candidate->serial, candidate->serialPort);
+			GreaseWeazleProto gwconfig = config.usb().greaseweazle();
+			return createGreaseWeazleUsb(candidate->serialPort, gwconfig );
 	}
 }
 
--- a/lib/usb/usbfinder.cc
+++ b/lib/usb/usbfinder.cc
@@ -45,5 +45,30 @@ std::vector<std::unique_ptr<CandidateDevice>> findUsbDevices(const std::set<uint
 		}
 	}
 
+    if (candidates.size() == 0) {
+      for (const auto& it : libusbp::list_connected_devices())
+        {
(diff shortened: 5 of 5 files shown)


davidgiven:
Hello --- thank you for this! 

Re the DTR change: I've ported this to Linux (and hopefully OSX as well --- we'll see). The only devices I've got are a FluxEngine and a real GreaseWeazle; I notice that toggling DTR doesn't reset the GreaseWeazle stream. Do you know if it's supposed to?

Re the serial port detection: I'm not actually keen on this implementation, as the serial ports are only tried if no FluxEngine or GreaseWeazle are plugged in. Plus, if you have one generic USB serial device on your machine then the client will try to talk the GreaseWeazle protocol to it which is liable to end up badly (e.g. if I have my 3D printer plugged in). I'll tinker with this.

(I suspect the CY8CKIT-059 is going to be EOL'd by Cypress, as it's out of stock everywhere, so alternative hardware is definitely required. I'd like to pivot the FluxEngine client towards being a general client for as many different flux-level readers as I can manage, so I'm definitely interested in supporting your hardware. However I'll note that the GreaseWeazle protocol does have some limitations, such as not supporting hard-sectored disks, so if your hardware has a native protocol I'd prefer to use that instead.)

ladyada:
hiya! the DTR doesn't reset the stream. with the CDC implementation we've got - if you dont set DTR the CDC driver just never sends data from the device to the host - that is, DTR is respected! it could be a 'bug' in TinyUSB but its also pretty harmless and [we found the DTR set code in a  minimal/boilerplate windows CDC example](https://stackoverflow.com/questions/6036716/serial-comm-using-writefile-readfile).

i agree that the generic serial method isnt great - note that we can't force a VID/PID in Arduino. greaseweazle project has a `--device` arg that allows specifying/forcing a COM port and we use that for gw. a similar technique would be great but we figured it would be a bigger PR so wanted to start with this instead. if ya want we can try to add `--device` or `--port`? 

right now we just have a script that emulates a gw device, we sorta like maintaining compatibility with existing projects instead of re-inventing everything - especially since y'all are clearly floppy experts :) for hard-sectors, we do want to add support for those at some point, since i've got a shugart 801... but maybe gw would be interested in adding a new command to handle them!

in this case, as the hardware for fluxengine (and anything with an stm32) is unobtainium, we wanted to do a generic arduino-ese port that would be less hardware dependent - i did find a cypress dev kit but its a bit $ these days. the library we've got currently supports RP2040 and SAMD51. do you have an RP2040 board of some kind, maybe a $4 pico board?

whew! so i guess, let me know if you want --port added and we could do that this weekend

davidgiven:
That's now merged, with some tweaks. To use GreaseWeazle mode with a manual serial port you can now do `--usb.greaseweazle.port=/dev/tty...` (or `COM...` on Windows). This bypasses the USB layer entirely. That seemed like the cleanest way to do things.

I've got some experience with the RP2040 --- I did the Fuzix port to it last year --- and it should be a decent machine for working with floppies. The IO processor looks ideal for timing flux pulses, which would let you DMA them straight into memory, freeing up the CPU to stream them via USB back to the host. There's probably also enough memory, just, to store a complete track in RAM (with the FluxEngine bytecode that's about 90kB for a IBM 1440kB disk) and enough CPU to do the decode on-board, opening up the possibility to use it as a generic mass storage device for lots of floppy disk formats. Which would be cool.

The bit that concerns me is that it's a 3.3V part. I picked the Cypress chip partly because it's the only 5V FPGA-ish microcontroller currently on the market that I know about. I know early versions of the GreaseWeazle made do without a level shifter, and it worked, but the RP2040 doesn't seem to be 5V tolerant. This is very much outside my comfort zone; I would expect there to be a current flow from the RP2040 to the drive if they both let a pin float high, but don't know whether this will cause any damage. (Another reason for using the Cypress chip is that I can solder the connector directly onto it without having to think about hardware!)

ladyada:
i dont think there's any pullups on floppy drives, so far i've had to enable the pullups on a board. pico's are also $4, they're...kinda disposable ;) its hard to find anything that is 5V-compliant that is inexpensive and fast these days. 
we have also created a demo of 'on the fly MFM-decoded' USB Mass Storage for 3.5/5.25 diskettes since you just have to offer up the blocks one by one. doing any other file format is, of course, a total nightmare and left as an exercise for the reader :)

https://github.com/adafruit/Adafruit_Floppy/blob/main/examples/msd_test/msd_test.ino

right now im fighting mac800k diskettes - all my disk drives just Do Not Like when the floppy has 10 or 11us pulses. very exciting!

davidgiven:
Some 5.25" drives have jumper-controlled pullups which can be as strong as 150R. keirf says these need buffering. https://github.com/keirf/Greaseweazle/wiki/Hardware-Setup#drive-terminations I'm not actually sure what the FluxEngine hardware has internally; I'll have to check (this might explain an issue someone reported). Modern drives use 10K. https://www.hermannseib.com/documents/floppy.pdf

Regarding Mac800k disks: yeah, have fun with GCR and the crazy checksum routine. I eventually gave up trying to understand the latter and just borrowed Mame's code. https://github.com/davidgiven/fluxengine/blob/master/arch/macintosh/decoder.cc

davidgiven:
Oh, yeah, and regarding long pulses, there's some FluxEngine tooling for analysing drive pulse response that might be useful. See http://cowlark.com/fluxengine/doc/driveresponse.html. If you're having problems you might want to try making sure you have a DD disk.

ladyada:
5.25" are definitely a different beast - we're planning to have buffered output for those. 
im going to dig up some proper DD disks, because my co-coder had no problems reading a commercial DD disk

PR #417, opened 2022-01-18
Add option to rescale flux source pulses.

tdaede:
This actually scales by the reciprotal of the flux sink, to
allow the same value to be set for both options.


davidgiven:
Merged. Thanks!

PR #422, opened 2022-01-22
Add NFD r0 image reader.

tdaede:
This format is very similar to D88, but used mostly for PC-98
rather than PC-88.


davidgiven:
So many weird image formats. Merged --- thanks!

PR #427, opened 2022-01-25
Improve d64 handling

jepler:
I was trying to use fluxengine d64 files with vice x64, but I found that when there were error sectors the disk image written out simply omitted them.

This caused vice to refuse to recognize the d64 file, because only a few specific file lengths are accepted. This made vice reject the floppy disk image with "unknown gcr version".

Instead, when a sector is not present that should have been, write out zeros and append an error bitmap.

I have used several pieces of software archived from floppy disks (using the adafruit prototype floppy featherwing) in x64 with this patch: Jumpman, Alf, and "Pro Financial Organizer" (hey, when you borrow whatever floppies a friend will lend you, you get weird stuff).

Screenshot_2022-01-24_20-41-47

### DIFF
--- a/lib/imagereader/d64imagereader.cc
+++ b/lib/imagereader/d64imagereader.cc
@@ -9,6 +9,14 @@
 #include <iostream>
 #include <fstream>
 
+enum {
+    SIZE_35T = 683*256,
+    SIZE_35T_ERRORS = 683*257,
+    SIZE_40T = 768*256,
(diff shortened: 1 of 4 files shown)


jepler:
This may need more work, I've found that it doesn't seem to append the error map when I expect it to.

davidgiven:
OK, let me know when you want me to have another look.

jepler:
I've totally lost the context on this by now. Closing.

PR #430, opened 2022-01-26
Support time-based reads for Greaseweazle V24+.

### LINKED ISSUE #426: Verification during write reads an excessive amount of flux with Greaseweazle
While writing a 360rpm disk:

  0.0: Write:   166 ms in 71541 bytes
       Verify:  326 ms in 140013 bytes

I would expect 166ms*1.2=199ms flux to be read. It seems that Greaseweazle does not support non-integer numbers of revolutions to be read. However, the code used to determine the integer # of revolutions also seems a bit suspect:

            int revolutions = (readTime+_revolutions-1) / _revolutions;

tdaede:
This substantially speeds up read and verify operations, allowing
only 1.2 revs rather than 2 revs to be read.

Fixes #426

### DIFF
--- a/lib/usb/greaseweazleusb.cc
+++ b/lib/usb/greaseweazleusb.cc
@@ -278,14 +278,13 @@ public:
         if (hardSectorThreshold != 0)
             Error() << "hard sectors are currently unsupported on the GreaseWeazel";
 
-        int revolutions = (readTime+_revolutions-1) / _revolutions;
-
         do_command({ CMD_HEAD, 3, (uint8_t)side });
 
         switch (_version)
         {
             case V22:
             {
+                int revolutions = (readTime+_revolutions-1) / _revolutions;
                 Bytes cmd(4);
                 cmd.writer()
                     .write_8(CMD_READ_FLUX)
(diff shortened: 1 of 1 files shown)


lib/usb/greaseweazleusb.cc
```diff
@@ -302,8 +301,8 @@ class GreaseWeazleUsb : public USB
                 cmd.writer()
                     .write_8(CMD_READ_FLUX)
                     .write_8(cmd.size())
-                    .write_le32(0) //ticks default value (guessed)
-                    .write_le32(revolutions + (synced ? 1 : 0));
+                    .write_le32((readTime + (synced ? _revolutions : 0))/ _clock)
```
davidgiven, line 304: _revolutions or revolutions?

davidgiven:
Fantastic!

Left one comment on something which looks suspicious to me.

lib/usb/greaseweazleusb.cc
tdaede: _revolutions (revolutions is only computed for V22)

davidgiven: OK. Not worth changing, but for reference I use nanoseconds_t for those to make it a bit clearer.

davidgiven: ...also there's a .clang-format in the root directory.

PR #432, opened 2022-01-26
Add flag to skip verification during write.

### LINKED ISSUE #425: Option to write without verify
This is a bit ill-advised but occasionally handy. It's already accessible by failing to specify the decoder format, but a real option would be more elegant, at least.

tdaede:
Fixes #425


davidgiven:
I was about to do this! Merged, thanks.

Issue #424, opened 2022-01-24
Image formats that specify floppy format don't verify if no format is set


tdaede:
e.g. DIM files verify when written with:

    ./fluxengine write ibm1232 -i file.dim

or:

    ./fluxengine write ibm -i file.dim

but not:

    ./fluxengine write -i file.dim

davidgiven:
This is because the code which is setting the format is only setting the encoder config. To perform the verification, it needs to set the decoder config too.

The fix should be trivial --- let me have a look.

PR #436, opened 2022-01-26
Fix some autoconfigure issues.

davidgiven:
- the DIM image reader now sets up the decoder, allowing writes to be verified
- the readers which do autoconfiguration will set the head and cylinder ranges where appropriate

Fixes: #424


tdaede:
Unfortunately this breaks the --no-verify option. That will have to be moved to be checked after the image is read.

davidgiven:
Yes, it will. I'll fix it. I've reopened #425.

PR #439, opened 2022-01-28
D88: Add 2D (40 track double density) support.

tdaede:
This assumes a 360rpm drive, and double stepping.

If fluxengine supports the concept of 40 vs 80 track drives in the
future, it might make sense for this to assume a 300rpm drive
instead.


davidgiven:
I believe this actually reverts a recent change I made, but I now notice that the change was wrong, so I'll fix it myself. Will merge. Thanks!

tdaede:
If you mean the automatic setting of has_cylinders()... D88 ended up setting it twice after your change. So I removed the other path that sets it (this allows it to be overridden also, previously D88 would always overwrite cylinder start and end unilaterally)

PR #447, opened 2022-02-08
fix 2 byte pack for revolution count

ladyada:
hiya, in greaseweazle, this is packed as a `H` (2 byte) not `I` (4 byte) value
https://github.com/keirf/greaseweazle/blob/ffdecb17bc5baf594921972c684999ff1d8eaf73/scripts/greaseweazle/usb.py#L396


davidgiven:
Yup, that won't help. Merged --- thanks.

PR #448, opened 2022-02-08
add dsk format which is basically diskcopy minus the 0x54 byte header

ladyada:
most mac emulators / computers dont like diskcopy format - they prefer dsk there's a converter but it would save a step if dsk's were emitted :)
see https://www.bigmessowires.com/dc2dsk.c
not sure if i did this right, and its a lil hacky, since we don't crop the file, but it does work!

image

### DIFF
--- a/lib/imagereader/dskimagereader.cc
+++ b/lib/imagereader/dskimagereader.cc
@@ -1,0 +1,135 @@
+#include "globals.h"
+#include "flags.h"
+#include "sector.h"
+#include "imagereader/imagereader.h"
+#include "image.h"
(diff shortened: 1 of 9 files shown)


ladyada:
btw mac400 works just fine too :)

image

image

davidgiven:
Actually, the existing image writer can handle these --- it just needs to be configured properly. I've done this and it's in the `macdsk` branch. PTAL and see if it works for you?

ladyada:
ooh didnt look there, ill fetch and check now

ladyada:
verified reading is good with that branch, testing writing next

PR #449, opened 2022-02-08
Change the mac profiles to produce simple sector images rather the DiskCopy files.

davidgiven:
DiskCopy 4.2 files are still available if you use a `.diskcopy` extension, as before.


ladyada:
reading and writing 800k and 400k tested!

ladyada (approved):

davidgiven:
Excellent, thanks!

PR #451, opened 2022-02-11
Support 84-track SCP files

jepler:
According to version 2.2 of the [SCP image specification at cbmstuff.com](https://www.cbmstuff.com/downloads/scp/scp_image_specs.txt) the number of Track Data Entries for a floppy can be up to 168, so that up to 84 tracks can be recorded (e.g., fluxengine -c 0-83).
    
> BYTES 0x10-0x2AF (for floppy image files) are a table of longwords with each entry being an offset to a Track Data Header (TDH) for each track that is stored in the image.  The table is always sequential.  There is an entry for every track, with up to 168 entries supported for floppy disks.  This means that floppy disk images of up to 84 tracks with sides 0/1 are possible.

Additionally, instead of segfaulting if a too-large number of tracks (e.g. -c 0-84) is used, print a message and continue.

### DIFF
--- a/lib/fluxsink/scpfluxsink.cc
+++ b/lib/fluxsink/scpfluxsink.cc
@@ -85,6 +85,12 @@ public:
 		trackdataWriter.seekToEnd();
 		int strack = strackno(cylinder, head);
 
+                if (strack >= std::size(_fileheader.track)) {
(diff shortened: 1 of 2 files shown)


davidgiven:
Much better --- thanks very much!

PR #459, opened 2022-02-17
Add formats for the BK.

davidgiven:


merclangrat:
I confirm that both work (bk800 and bk800_525) with 720/800Kb diskettes, both reading and writing, it can be merged. But, if Fluxengine doesn't need _525 versions anymore, it may need to be changed, right?

davidgiven:
Excellent news! Yes, the profiles will need adjusting for the new scheme, but I'll do that.

PR #461, opened 2022-02-19
upgrade-flux-file: Fix for Windows

hharte:
The upgrade-flux-file utility does not work properly under Windows.  It runs into an error because the open database file cannot be renamed in-place.

Close the database and remove the input file prior to renaming.

Tested this change on Windows, MacOS, Linux.


davidgiven:
Bah. Thanks; merged.

Issue #328, opened 2021-08-28
Micropolis decoder struggling


cube1us:
The Micropolis hard sector decoder is only correctly decoding about 1/3 of my Mod I disks (35 tracks, 48 TPI).  I haven't tried the Mod II's much yet, but from what I did try, I suspect it will struggle there as well.

If you'd like me to capture some flux files etc. let me know what parameters.  I have flux files for them all, but they may not have quite what you want.

ejona86:
I would be interested in some flux files to improve the support. For parameters, if you are using greaseweazle, then at least 20 revolutions to get 1.25 "real" revolutions. If you already have flux files, then they are likely fine as-is. 1-2 flux files is probably enough, but obviously I can verify more of them if I have more.

One thing that would be helpful is to know what machine you are using the Mod I with. The host has a lot of say in how they were written. I have come across some (well, at least one) VGI files that don't seem to use the normal checksum, so I suspect there may have been some variation present. It is interesting that 1/3 of your disks don't work well; I would have suspected it to be more consistent than that. Do you think there is anything special about those 1/3 of the disks that make them different from the other 2/3?

cube1us:
See issue https://github.com/davidgiven/fluxengine/issues/322 for a 40 revolution SCP file for a Mod II.  [Or, do you mean something else by "flux file"??].    In my case, ALL of both the Mod I and Mod II disks were written using a Micropolis controller on a 2MHz 8080 (Altair), mostly under CP/M.  I doubt that there is anything special about the different disks.  (The only difference between MOD I and MOD II is TPI - density and recording mode is the same.)

ejona86:
@cube1us, how does this issue differ from #322? Since I demonstrated that once the SCP import is working properly the Micropolis decoder seemed to work fine, is there anything more to do here and instead just let #322 track the SCP issues?

cube1us:
#322 was for a .scp file created using greaseweazle to read the floppy - in that case it would not process a .scp file AT ALL.      THIS issue was after I got a fluxengine hardware and tried to read Micropolis floppies with it -- no .scp files were involved.  The results were/are suspicious because so many floppies had errors and would not successfully decode.  I had successfully read almost all of these floppies 16 years ago just fine using my Altair.  And I have had similar success with non-Micropolis floppies with my Greaseweazle this year, and all of these floppies were stored similarly, which led me to conclude that the decoder was not working as well as one might hope, since this issue relates to .flux files read with fluxengine hardware.

[...]

PR #463, opened 2022-02-19
Improve micropolis decoder

hharte:
This partially addresses Issue #328.

Since the decoders are not allowed to seek anymore, after finding a SYNC that is too early after the sector pulse, search for 00's to advance, and then search for a SYNC again.

I'm not sure if it's the disks that I have (ie, possibly old/damaged) or whether the MFM decoder could be improved, but the decoder still struggles on some disks.

FYI: @ejona86


davidgiven:
LGTM. Thanks!

hharte:
Thanks David,

I believe there may still be some improvements possible in the MFM decoder.  Some of the errors I've seen look like they are a single flux transition off.  Do you have any ideas about how to go about debugging this?

PR #468, opened 2022-02-22
Fix various doc typos

JohnVeness:
Hi, I have fixed a few typos in documentation files only. Thanks :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ FluxEngine
 ==========
 
 (If you're reading this on GitHub, the formatting's a bit messed up. [Try the
-version on cowlark.com instead.](http://cowlark.com/fluxengine/)
+version on cowlark.com instead.](http://cowlark.com/fluxengine/))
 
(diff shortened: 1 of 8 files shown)


JohnVeness:
There is also something wrong in disk-c64.md in the line:
`1581 disks are just another version of the standard B Just do:`
But I'm not quite sure what the correct text is.

davidgiven:
Wow --- that embarrassing, particularly as my mother was an English teacher... let me know when it's ready and I'll merge. Thanks!

JohnVeness:
Haha, no worries! It's mostly very good, I'm just nit-picking really :)

I think I'm done for the time being, other than noticing that on cowlark.com there is an oddity where an extra Acorn ADFS entry appears in the navigation, but I can't see where to fix that.
image

davidgiven:
The last will be an issue with my CMS --- will fix. Thank you very much for doing this!

PR #471, opened 2022-02-23
Add support for PDP-11 RX-50 disks

jepler:
According to my source [https://www.cbmstuff.com/forum/showthread.php?tid=634] the format of RX-50 is

 * single sided
 * 80 tracks
 * 10 sectors per track
 * 96 tpi (tracks per inch)
 * 300 rpm (revolutions per minute)
 * 250 KHz data rate

I have a disk labeled
```
BL-T540E-MC
CZUFDE0 MICRO PDP-11
USER TESTS
© 1983, 1894
DIGITAL EQUIPMENT CORP.
```
and stamped 14131.

The image read all sectors and the content looks plausible.  `strings` on it says things like
```
$ strings rx50.img  | grep -i pdp
; This file implements the USER FRIENDLY Diagnostic for the MICRO PDP11 & 11/73
```
however, I don't have an emulator or other system to use the disk image with so
I can't 100% vouch for the image being complete and correct.

### DIFF
--- a/src/formats/rx50.textpb
+++ b/src/formats/rx50.textpb
@@ -1,4 +1,4 @@
-comment: 'Digital RX50 400kB 5.25" 80-track 10-sector SSHD'
+comment: 'Digital RX50 400kB 5.25" 80-track 10-sector SSQD'
 
 flux_sink {
 	drive {
--- a/src/formats/rx50.textpb
+++ b/src/formats/rx50.textpb
@@ -1,0 +1,76 @@
+comment: 'RX50 450kB 5.25" 80-track 10-sector SSHD'
+
+flux_sink {
+	drive {
+		high_density: true
+	}
+}
+
+flux_source {
+	drive {
+		high_density: true
+	}
+}
+
+image_reader {
+	filename: "ibm1200_525.img"
+	img {
+		tracks: 80
+		sides: 2
+		trackdata {
+			sector_size: 512
+			sector_range {
+				start_sector: 1
+				sector_count: 10
+			}
+		}
+	}
+}
+
+image_writer {
+	filename: "ibm1200_525.img"
+	img {}
+}
+
+encoder {
+	ibm {
+		trackdata {
+			track_length_ms: 167
+			clock_rate_khz: 500
+			sectors {
+				sector: 1
+				sector: 2
+				sector: 3
+				sector: 4
+				sector: 5
+				sector: 6
+				sector: 7
+				sector: 8
+				sector: 9
+				sector: 10
+			}
+		}
+	}
+}
+
+decoder {
+	ibm {
+		trackdata {
+			sector_range {
+				min_sector: 1
+				max_sector: 10
+			}
+		}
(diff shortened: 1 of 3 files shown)


src/formats/rx50.textpb
```diff
@@ -0,0 +1,76 @@
...
+			sector_range {
+				start_sector: 1
+				sector_count: 10
+			}
+		}
+	}
+}
+
+image_writer {
+	filename: "ibm1200_525.img"
```
hharte, line 31: filename: "rx50.img"

davidgiven:
Excellent, thanks! Yes, please change the image_reader and image_writer filenames. Also, I think I probably want to put together a generic PDP-11 page.

jepler:
Here's another source of info about RX50: http://home.claranet.nl/users/pb0aia/vax/rx50.html

should I put
```
                        track_length_ms: 200
                        clock_rate_khz: 250
```
or should I do like you have for ibm360_525 and use 167/300? The current values are wrong, in any case.

jepler:
as for doc, since this is essentially a variant on "generic IBM" should I just add a bullet to disk-ibm.md instead?

davidgiven:
Re track length: that should be 167/300, because it's referring to the rotation period and data rate in a normal 360rpm PC drive. This all needs fixing to be independent of the drive type, as it means you can't write disks in 300rpm drives if you happen to have one, but that's where we are now.

Re doc: add a bullet to disk-ibm.md. I'll update things later.

Thanks for the link!

davidgiven:
Also, from looking at that link... isn't the RX-50 a Vax drive rather than a PDP-11 one?

jepler:
I don't really know much about this era of computing; I have no recollection of why I even have these floppies.

The RX50 is listed as a PDP-11 accessory by wikipedia https://en.wikipedia.org/wiki/PDP-11#Peripherals and the label on my disks says PDP-11 and RX50K. Here's an image of one of them:
image

jepler:
@davidgiven review suggestions implemented & pushed.

hharte (approved):

davidgiven:
Actually, one more request: in the mkninja.sh file, down the bottom, there's a block of `encodedecodetest` lines. Could you add the rx50 format there? That will run a basic random-data-to-flux-and-back-to-data image to make sure a round trip conversion works. Other than that, it all looks LGTM to me.

jepler:
Thanks for the prompt to add a test. I discovered that the round-trip didn't work, and resolved it by reducing the 'gap3' to match a value I found in flashfloppy. I wasn't able to find a different source for this value, the closest I came https://archive.org/details/bitsavers_sigmaInfor0FloppyCtrlManJun87_1225619/page/n17/mode/2up table didn't give a specific number of bytes in the table, just 'P'.

davidgiven:
Thanks --- merged!

jepler:
Thank you both for all the advice during this PR. Happy to see it merged, hope it helps someone in the future.

PR #473, opened 2022-02-25
fluxmap: Don't push_back() if Fluxmap begins with F_DESYNC.

hharte:
Fluxmap::split() creates a 0-length Fluxmap if the Fluxmap
begins with F_DESYNC.  Fix this by not doing push_back()
if a F_DESYNC is encountered at the start of the Fluxmap.


davidgiven:
Yup --- merged. Ta.

PR #475, opened 2022-02-26
Improve hard sector decode

hharte:
seekToPattern() can skip the index hole if it doesn't find the
SYNC pattern.  If this happens too close to the end of the flux
stream, it can result in a conflicted sector.  In this case,
discard the sector.

This makes a small improvement for #328 on some disks.


davidgiven:
Merged, thanks. One day I need to find some hard-sectored disks to try this for myself...

PR #477, opened 2022-02-27
WIP: Process flux segments separately.

hharte:
When a physical disk is read and retries are necessary, the
retried track's flux is concatenated in the .flux file, separated
by F_DESYNC delimiters.  The reader currently processes the single
multi-segment flux, leading to differences in interpretation of
a physical disk read vs. processing the .flux file generated by
the same physical disk read.

Break up the Fluxmap and process each segment separately.

An alternate approach would be to let readFluxmap() return the flux up until the F_DESYNC, and then let the reader make subequent calls to readFluxmap() returning the next segment of flux from the same track as part of the retry mechanism.

I'd like to get feedback on these two approaches.  The goal is for .flux files processed with "fluxengine read" to be processed identically to reading a physical disk.

### DIFF
--- a/lib/reader.cc
+++ b/lib/reader.cc
@@ -110,39 +110,49 @@ void readDiskCommand(
                 auto fluxmap = readFluxmap(fluxsource, cylinder, head);
                 totalFlux.appendDesync().appendBytes(fluxmap->rawBytes());
 
-                auto trackdataflux =
-                    decoder.decodeToSectors(fluxmap, cylinder, head);
-                track->trackDatas.push_back(trackdataflux);
+                auto maps = fluxmap->split();
 
-                track_sectors.insert(trackdataflux->sectors.begin(),
-                    trackdataflux->sectors.end());
-                track_records.insert(trackdataflux->records.begin(),
-                    trackdataflux->records.end());
-
+                std::shared_ptr<const TrackDataFlux> trackdataflux;
                 bool hasBadSectors = false;
-                std::set<unsigned> required_sectors =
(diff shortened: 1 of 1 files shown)


davidgiven:
Originally my plan was to have multiple fluxmaps for successive reads, but that caused problems due to design impedance mismatch with things like storing multiple fluxmaps in the flux files. But that predates the fl2 change, so maybe it's worth revisiting that. I agree that having flux files and hardware behave similarly is valuable. I'm quite willing to believe that F_DESYNC is a bad idea.

The biggest knock-on change there is that FluxSource::readFluxmap would probably need to return a kind of iterator rather than returning a single fluxmap directly. That wouldn't be hard, TBH. I'll think about this one.

davidgiven:
After spending some time poking at the code: I remember now why I gave up on this. You end up with two types of retryable flux sources: those which can reread indefinitely, and those which have a fixed number of retries. This becomes important for rawread and rawwrite where you may want to copy all available flux if the target is a file.

I believe the latter's broken anyway. With rawread, you only ever get one read because no error detection is done. And with rawwrite, all data _including the desync bytes_ is sent to the hardware, which is wrong; but rawwrite only works with solved fluxmaps anyway.

I think I have a route to fixing this properly...

davidgiven:
Okay, so #488 contains a big refactor to store multiple fluxmaps throughout, hopefully making F_DESYNC go away completely. It compiles and the tests pass...

hharte:
Awesomeness me, thank you!

PR #480, opened 2022-02-28
Add apple2 encoder

jepler:
This is tested with encodedecodetest.sh but is not tested on HW yet. It's likely that the sector order (interleave) doesn't match real systems.

### DIFF
--- a/arch/apple2/encoder.cc
+++ b/arch/apple2/encoder.cc
@@ -1,0 +1,187 @@
+#include "globals.h"
+#include "arch/apple2/apple2.h"
+#include "decoders/decoders.h"
+#include "encoders/encoders.h"
+#include "sector.h"
+#include "writer.h"
+#include "image.h"
+#include "fmt/format.h"
+#include "lib/encoders/encoders.pb.h"
+#include <ctype.h>
+#include "bytes.h"
+
+static int encode_data_gcr(uint8_t data) {
+    switch(data)
+    {
+	#define GCR_ENTRY(gcr, data) \
+	case data: return gcr;
+	#include "data_gcr.h"
+	#undef GCR_ENTRY
+    }
+    return -1;
+}
+
+class Apple2Encoder : public AbstractEncoder
+{
+public:
+    Apple2Encoder(const EncoderProto& config):
+	AbstractEncoder(config)
+    {}
+
+public:
+    std::vector<std::shared_ptr<const Sector>> collectSectors(int physicalTrack, int physicalSide, const Image& image) override
+    {
+	std::vector<std::shared_ptr<const Sector>> sectors;
+	constexpr auto numSectors = 16;
+	if (physicalSide == 0)
+	{
+	    int logicalTrack = physicalTrack / 2;
+	    unsigned numSectors = 16;
+	    for (int sectorId=0; sectorId<numSectors; sectorId++)
+	    {
+		const auto& sector = image.get(logicalTrack, 0, sectorId);
+		if (sector)
+		    sectors.push_back(sector);
+	    }
+	}
+
+	return sectors;
+    }
+
+    std::unique_ptr<Fluxmap> encode(int physicalTrack, int physicalSide,
+	    const std::vector<std::shared_ptr<const Sector>>& sectors, const Image& image) override
+    {
+	if (physicalSide != 0)
+	    return std::unique_ptr<Fluxmap>();
+
+	int logicalTrack = physicalTrack / 2;
+	double clockRateUs = 4.;
+
+	int bitsPerRevolution = 200000.0 / clockRateUs;
+
+	std::vector<bool> bits(bitsPerRevolution);
+	unsigned cursor = 0;
+
+	for (const auto& sector : sectors) {
+	    if(sector) {
+		writeSector(bits, cursor, *sector);
+	    }
+	}
+
+	if (cursor >= bits.size())
+	    Error() << fmt::format("track data overrun by {} bits", cursor - bits.size());
+	fillBitmapTo(bits, cursor, bits.size(), { true, false });
+
+	std::unique_ptr<Fluxmap> fluxmap(new Fluxmap);
+	fluxmap->appendBits(bits, clockRateUs*1e3);
+	return fluxmap;
+
+    }
+
+private:
+    uint8_t volume_id = 254;
+
+    /* This is extremely inspired by the MESS implementation, written by Nathan Woods
+     * and R. Belmont: https://github.com/mamedev/mame/blob/7914a6083a3b3a8c243ae6c3b8cb50b023f21e0e/src/lib/formats/ap2_dsk.cpp
+     * as well as Understanding the Apple II (1983) Chapter 9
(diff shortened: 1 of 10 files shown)


jepler:
Feedback on style & correctness is appreciated. I hope to test this on hardware within the week, but no guarantees.

src/formats/apple2.textpb
```diff
@@ -1,5 +1,20 @@
 comment: 'Apple II 140kB DOS 3.3 5.25" 40 track SSSD (ro)'
```
davidgiven, line 1: Not (ro) any more!

arch/apple2/encoder.cc
```diff
@@ -0,0 +1,187 @@
...
+	fillBitmapTo(bits, cursor, bits.size(), { true, false });
+
+	std::unique_ptr<Fluxmap> fluxmap(new Fluxmap);
+	fluxmap->appendBits(bits, clockRateUs*1e3);
+	return fluxmap;
+
+    }
+
+private:
+    uint8_t volume_id = 254;
```
davidgiven, line 82: Is this ever likely to need to change? Is it worth putting it in the proto file as a parameter?

davidgiven:
Thank you very much --- I left a couple of nitpick comments, but other than that I reckon it's good to go. I would ask that you also update `doc/disk-apple2.md` to mention writing. You might also like to put a dinosaur in the Apple II Write column in the table in the README.

(I'm not particularly precious about style given that I have trouble maintaining a consistent style myself, but there's a .clang-format in the project root.)

jepler:
I tried this on hardware and .. no dice yet.

I have a just-received pair of double sided floppies from TheApple2OnlineStore via ebay. Each side of the disk either has DOS 3.3 or ProDos. All 4 sides boot just dandy in my Apple //e, so I know they work.  However, (before this change) `fluxengine read apple2` reports errors reading virtually all sectors of any side of the disks, using a genuine GreaseWeazle and a TEAC floppy drive. This combination is known to read and write write good .d64 files for Commodore 64 so it should not have any trouble operating with the Apple II formats (in terms of problems with drive density settings, flux lengths, etc) Does anyone else know the state of Apple II _reading_ in FluxEngine, or when someone else was last using it successfully?

I can write and verify `ProDOS Users' Disk` (https://archive.org/details/ProDOS_Users_Disk) with FluxEngine but it doesn't boot. There's _something_ there, as `CATALOG` from DOS3.3 can print "DISK VOLUME 254" before it gets "I/O ERROR" (not totally unexpected, as the content is meaningless to DOS3.3), and attempting to boot from it drops into the monitor at $0914 on a BRK instruction. The content of the boot sector starting at $0800 looks right in memory (800L in the monitor to disassemble). The content starting at $0900 is wrong. This may be down to the sector ordering within the input image file, though.

I also tried writing the DOS3.3 image and it also doesn't boot (eventually crashes out to monitor) but it can be CATALOG'd, though actually loading/running the files doesn't work either. So, something's working but not everything. I don't know about the apparently pre-existing problem on the reading side, but the problems on the writing side may have to do with me being unclear about the order in which Apple 2 images are stored. I'll have to do more reading about that.

I'll keep experimenting, and also work on the suggestions so far given in this PR. Thanks!

hharte:
You may want to try checking out FluxEngine code before this commit about 20 days ago: 0933dc1afa27aa288491f37bede9639f61744a04

and see if Apple ][ works.  IIRC, a few of the decoders were broken by that change, particularly the hard-sectored ones, because they do not work with the encodedecode tests.

I don’t have any Apple disks to try, but I do have an Apple ][ computer, so I’d love to see writing working so I can make some disks for it :)

davidgiven:
Addendum to the last: as un-remapped disk images are no use to man nor dog, this needs to be fixed. The easiest thing is for me to add support to the image writer to rearrange sectors and then use the new modifier profile support to configure it, so you'd do something like `fluxengine read apple2 prodos -s drive:0` or whatever. That ought to be relatively straightforward to implement.

jepler:
I'm working with 6-and-22 disks (DOS 3.3 and ProDOS) so that's not an issue.

What I'm seeing in the decoder with these boot disks from ebay is that after the APPLE2_DATA_RECORD mark, there's an additional 0 bit before the valid GCR data starts. Snarfing this bit if it's a 0 make almost the whole DOS 3.3 disk decode successfully, albeit in a weird order. I think this extra `0` bit would go unnoticed on the real hardware, as the shift register is read out only when a `1` is in its MSB. OTOH, I could be all wrong with my diagnosis.

```c
               if (readRaw24() != APPLE2_DATA_RECORD)
                       return;

                // Sometimes there's a gap between APPLE2_DATA_RECORD and the data itself!
                // this is verified on floppies written on a real Apple //e that is otherwise
                // believed to be in working condition.
                //
                // This would not affect the real disk hardware, as it was a '1' reaching the
                // top bit of a shift register that triggered SW, but it affects the way the data
                // is read here.
                auto p = tell();
                auto b = readRawBits(1);
                if(b[0]) seek(p);
```

When I write a disk image (DOS 3.3) read by FluxEngine with this modification this to a fresh floppy, I can at least CATALOG the whole thing, though I still failed to boot it or to run any programs. However, the FluxEngine read DID have some CRC errors in the final result still. I should try a different piece of media.

To me, this confirms the idea that one reason writing a DOS or ProDOS image from the internet isn't working is that FluxEngine isn't taking account of the sector numbering. But at this point I don't feel confident enough to say whether this is the only problem, because otherwise things would be working better.

I'd prefer to avoid having this PR be "fix everything about apple2 reading and writing". Can we define what would be an end-point to allow this PR to be merged, and then continue working on it in future PRs? Perhaps doing the doc and style things mentioned above, adding sector remapping on write (I think there are 3 settings? Identity (in case anyone has a disk read by FluxEngine before this PR), DOS3.3 and ProDos?), and proving that the resulting disk can be read on my //e?

davidgiven:
I think that if it round-trips through encodedecodetest, it's good enough to merge. After all, it's not any _more_ broken than it was previously...

I'll do the remapping support, and then let's continue with another PR.

Thank you for testing this; without real hardware it's really hard to know what works and what doesn't.

arch/apple2/encoder.cc
jepler: It may need to be. But, can we leave it to the future?

I have a limited selection of disks to check, but DOS 3.3 and a disk created from it with `INIT` both come up VOLUME 254. However, ProDOS comes up as VOLUME 1 (when listed from inside DOS 3.3 `CATALOG` which shows it; ProDOS doesn't seem to show the volume number).

For DOS 3.3 the volume number can be detected in the VTOC track 17 sector 0 byte 6. It can also be specified to INIT:
```
Each DOS 3.3 diskette has a Volume Number set at the time it is initialized. The
default Volume Number used by INIT is 254. You can set the number to something
else at the time a diskette is initialized:

INIT HELLO, V19```
for prodos I didn't find how to format with a different volume ID than 1, though a disk converted from DOS 3.3 _to_ ProDOS may retain its original volume ID.

davidgiven: Yes, no problem.

PR #481, opened 2022-03-01
apple2: Comment decoder, add fix/workaround for stray 0 at start of sector data

jepler:
Sometimes there's a 1-bit gap between APPLE2_DATA_RECORD and the data itself.  This has been seen on real world disks such as the Apple II Operating System Kit from Apple2Online.  However, I haven't seen it described in any of the various references.                    
                  
This extra '0' bit would not affect the real disk interface, as it was a '1' reaching the top bit of a shift register that triggered a byte to be available, but it affects the way the data is read here.                                  

With this change, and a separate program to manipulate an image between physical order and logical DOS3.3 order, I was able to verify that the "Apple II Operating System Kit Apple DOS 3.3" disk that I imaged with FluxEngine is 1 bit different from the image available at https://apple2online.com/?page_id=4262 `Apple-DOS-All-Versions-1/Apple DOS 3.3 August 1980.dsk`. The difference is in a byte in the VTOC (filesystem metadata) that is marked as unused in the references I have. I don't know why this difference exists.

However, it feels like reading apple2 images is still pretty marginal, it took 9 separate read attempts (not the retry process within fluxengine, which was set at the default of 5 during this process) before I got an error-free disk image. A floppy freshly `INIT`ialized from DOS 3.3 is no better, and may be worse. I'll continue investigating on my end, because I don't know if this is the end of the story.

I also added some comments about how the sector validity flag is handled through the read process; I hope it's an improvement over simply announcing that the code is `/* unintuitive but correct */`.

### DIFF
--- a/arch/apple2/decoder.cc
+++ b/arch/apple2/decoder.cc
@@ -104,6 +104,20 @@ public:
 		if (readRaw24() != APPLE2_DATA_RECORD)
 			return;
 
+		// Sometimes there's a 1-bit gap between APPLE2_DATA_RECORD and
+		// the data itself.  This has been seen on real world disks
+		// such as the Apple II Operating System Kit from Apple2Online.
+		// However, I haven't seen it described in any of the various
+		// references.
+		//
+		// This extra '0' bit would not affect the real disk interface,
+		// as it was a '1' reaching the top bit of a shift register
+		// that triggered a byte to be available, but it affects the
+		// way the data is read here.
+		auto p = tell();
+		auto b = readRawBits(1);
+		if(b[0]) seek(p);
+
 		/* Read and decode data. */
 
 		unsigned recordLength = APPLE2_ENCODED_SECTOR_LENGTH + 2;
--- a/arch/apple2/decoder.cc
+++ b/arch/apple2/decoder.cc
@@ -90,6 +90,9 @@ public:
 		_sector->logicalTrack = combine(br.read_be16());
 		_sector->logicalSector = combine(br.read_be16());
 		uint8_t checksum = combine(br.read_be16());
+
+		// If the checksum is correct, upgrade the sector from MISSING
+		// to DATA_MISSING in anticipation of its data record
 		if (checksum == (volume ^ _sector->logicalTrack ^ _sector->logicalSector))
 			_sector->status = Sector::DATA_MISSING; /* unintuitive but correct */
 	}
@@ -106,6 +109,9 @@ public:
 		unsigned recordLength = APPLE2_ENCODED_SECTOR_LENGTH + 2;
 		Bytes bytes = toBytes(readRawBits(recordLength*8)).slice(0, recordLength);
 
+		// Upgrade the sector from MISSING to BAD_CHECKSUM.
+		// If decode_crazy_data succeeds, it upgrades the sector to
+		// OK.
 		_sector->status = Sector::BAD_CHECKSUM;
 		_sector->data = decode_crazy_data(&bytes[0], _sector->status);
 	}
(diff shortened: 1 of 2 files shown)


arch/apple2/decoder.cc
```diff
@@ -101,11 +104,28 @@ class Apple2Decoder : public AbstractDecoder
...
+		// However, I haven't seen it described in any of the various
+		// references.
+		//
+		// This extra '0' bit would not affect the real disk interface,
+		// as it was a '1' reaching the top bit of a shift register
+		// that triggered a byte to be available, but it affects the
+		// way the data is read here.
+		auto p = tell();
+		auto b = readRawBits(1);
+		if(b[0]) seek(p);
```
davidgiven, line 119: You can't do seek() here. The flux decoder has internal state which gets reset when a seek happens (or, more accurately, _should_ be reset when a seek happens and frequently isn't). I had to do a major refactor a few weeks ago to remove spurious seeks which were corrupting the bit stream.

It's probably possible to do this by fiddling with the flux decoder's state, but it might be easier to just read recordLength*8+1 bits and discard the first if it's a zero. There should be a gap afterwards so the extra bit shouldn't do any harm.

arch/apple2/decoder.cc
```diff
@@ -101,11 +104,28 @@ class Apple2Decoder : public AbstractDecoder
...
+
 		/* Read and decode data. */
 
 		unsigned recordLength = APPLE2_ENCODED_SECTOR_LENGTH + 2;
 		Bytes bytes = toBytes(readRawBits(recordLength*8)).slice(0, recordLength);
 
+		// Upgrade the sector from MISSING to BAD_CHECKSUM.
+		// If decode_crazy_data succeeds, it upgrades the sector to
+		// OK.
```
davidgiven, line 128: That's an improvement; thanks.

arch/apple2/decoder.cc
jepler: I see, thanks. I'll reformulate it.

jepler: Do you feel any better about doing it like this?
```c++
                // Sometimes there's a 1-bit gap between APPLE2_DATA_RECORD and
                // the data itself.  This has been seen on real world disks
                // such as the Apple II Operating System Kit from Apple2Online.
                // However, I haven't seen it described in any of the various
                // references.
                //  
                // This extra '0' bit would not affect the real disk interface,
                // as it was a '1' reaching the top bit of a shift register
                // that triggered a byte to be available, but it affects the
                // way the data is read here.
                //
                // While the floppies tested only seemed to need this applied
                // to the first byte of the data record, applying it
                // consistently to all of them doesn't seem to hurt, and
                // simplifies the code.
                
                /* Read and decode data. */   
                
                auto readApple8 = [&]() {
                    auto result = readRaw8();
                    while(result & 0x80 == 0) {
                        auto b = readRawBits(1);
                        result = (result << 1) | b[0];
                    }
                    return result;
                };
 
                constexpr unsigned recordLength = APPLE2_ENCODED_SECTOR_LENGTH + 2;
                uint8_t bytes[recordLength];
                for(auto &byte : bytes) {
                    byte = readApple8();
                }
```

davidgiven: Yes, that looks fine. Will merge.

davidgiven:
Merging --- thanks!

I have add sector remapping support. It's currently in https://github.com/davidgiven/fluxengine/tree/mapper. There's a prototype AppleDOS mapping table based on some numbers what I found on the internet, so I have no idea whether they're right or whether the mapper even works. Any chance you could take a look?

To use, do:

`fluxengine read appleii appledos -s ...usual stuff here...`

It should work (in reverse) for writing too.

jepler:
Thank you! It'll likely be next week before I can test it, though. Please feel free to give me a ping if I drop the ball.

PR #489, opened 2022-03-08
Further apple fixes - have now booted a DOS 3.3 written with FE + GW

jepler:
With this change, I was able to boot a DOS 3.3 disk I rearranged into "physical" order; some 'there's a disk error' raspberries occur (probably indicating a data error within a sector) during the boot process, so something is still obviously marginal, but this is a huge step forward.

It also fixes a crash that would occur reading an apple2 disk that was not aligned to the index sensor; during my testing, I was using an image I had written myself, so of course it was aligned. (At least, I think this is why I even believed it worked at all)


davidgiven:
Merged.

PR #490, opened 2022-03-08
Upgrade Apple II writing to 'unicorn', fix & explain remapping

jepler:
* Update documentation (note apple flux writing is a unicorn now even though this isn't quite true until #489)
 * Fix DOS 3.3 mapping
 * Add ProDOS (all versions) mapping


davidgiven:
Thank you!

PR #492, opened 2022-03-10
Add support for the interfacing to an Apple Disk ][

jepler:
Some folks may want to use interfaces besides PC/Shugart ones. Together with https://github.com/adafruit/Adafruit_Floppy/pull/15 and a hand wired board I'm able to read flux from an Apple Disk ][ (don't sweat the bad sectors, this floppy was written before I mostly got the bugs out of flux generating for apple2s)

I've also started a dialog with @kierf about allocating these disk-type numbers. However, we haven't settled the details yet. This actual working code may be useful in figuring out the way forward among all of us:
 * https://github.com/keirf/greaseweazle/pull/171

Writing will have an additional wrinkle: The Apple2 stepping mechanism is designed so that it's possible to reach 3 intermediate positions between each of the ~35 data tracks. Because we hope to image protected floppies, which in some cases are said to  se these "quarter tracks", we want to expose the ability to reach all ~140 positions. This is APPLE2_QUARTERSTEP. However, right now the code to transform from a disk image's "cylinder number" to a floppy mechanism's "cylinder number" is a bit spread across the source. This may mean that it's just easiest to use APPLE2 (which presents 1 intermediate position between data tracks, just like when you stick an Apple disk in a PC drive) when writing.

In any case, I didn't wire the write and write-protect lines yet so writing isn't tested at all. The electrical interface is expected to be just fine.

```
$ ./fluxengine read apple2 --usb.greaseweazle.bus_type=APPLE2_QUARTERSTEP --usb.greaseweazle.port=/dev/ttyACM1 --flux_source.drive.high_density=0 -c 0-140x4 --decoder.retries=1
      38 raw records, 19 raw sectors; 4.04us clock (247kHz)
      sectors: 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
      logical track 35.0
      4096 bytes decoded
     Tracks -> 1         2         3
H.SS 012345678901234567890123456789012345
0. 0 ....................................
0. 1 ....................................
0. 2 .BC...........CC...............C....
0. 3 ....................................
0. 4 ....................................
0. 5 ....................................
0. 6 B..B................................
0. 7 B...................................
0. 8 ....................................
0. 9 ....................................
0.10 B...................................
0.11 ....................................
0.12 ....................................
0.13 ....................................
0.14 ....................................
0.15 ....................................
Good sectors: 567/576 (98%)
Missing sectors: 0/576 (0%)
Bad sectors: 9/576 (1%)
      IMG: wrote 36 tracks, 1 sides, 144 kB total
```

```
$ ./fluxengine inspect --usb.greaseweazle.bus_type=APPLE2 --usb.greaseweazle.port=/dev/ttyACM1 --flux_source.drive.high_density=0
      Using GreaseWeazle on serial port /dev/ttyACM1
Measuring rotational speed... 202.3ms (296.6rpm)
0x12328 bytes of data in 242.003ms
Required USB bandwidth: 300.777210636634kB/s

Clock detection histogram:
 47 3.92    157 █
 48 4.00   1773 ███████████▋
 49 4.08   3273 █████████████████████▌
 50 4.17   2171 ██████████████▎
 51 4.25   2727 █████████████████▉
 52 4.33   3423 ██████████████████████▌
 53 4.42   1072 ███████
 54 4.50     93 ▌
...
 68 5.67    284 █▊
 69 5.75   2503 ████████████████▍
 70 5.83   4926 ████████████████████████████████▍
 71 5.92   5812 ██████████████████████████████████████▏
 72 6.00   6082 ████████████████████████████████████████
 73 6.08   3950 █████████████████████████▉
 74 6.17   3539 ███████████████████████▎
 75 6.25   1959 ████████████▉
 76 6.33    253 █▋
...
 94 7.83     61 ▍
 95 7.92    107 ▋
 96 8.00     76 ▍
 97 8.08     45 ▎
 98 8.17     55 ▎
 99 8.25     34 ▏
...
Noise floor:  60
Signal level: 304
Peak start:   46 (3.83 us)
Peak end:     55 (4.58 us)
Median:       50 (4.17 us)
4.17 us clock detected.
```

### DIFF
--- a/doc/greaseweazle.md
+++ b/doc/greaseweazle.md
@@ -37,8 +37,12 @@ Supported features with the GreaseWeazle include:
   - simple reading and writing of disks, seeking etc
   - erasing disks
   - determining disk rotation speed
-  - both Shugart and normal IBM buses (via
+  - Shugart and normal IBM buses (via
 	`--usb.greaseweazle.bus_type=SHUGART` or `IBMPC`; the default is `IBMPC`)
+  - Apple 5.25 floppy interfaces (via `--usb.greaseweazle.bus_type=APPLE2`)
+
+Which device types are supported depend on the hardware. Genuine Greaseweazle hardware supports SHUGART and IBMPC.
+APPLE2 is only supported with hand wiring and the Adafruit\_Floppy greaseweazle-compatible firmware.
 
 What doesn't work
 -----------------
--- a/lib/usb/greaseweazle.h
+++ b/lib/usb/greaseweazle.h
@@ -75,9 +75,10 @@ extern Bytes stripPartialRotation(const Bytes& fldata);
 /*
  * CMD_SET_BUS CODES
  */
-#define BUS_NONE            0
-#define BUS_IBMPC           1
-#define BUS_SHUGART         2
+#define BUS_NONE              0
+#define BUS_IBMPC             1
+#define BUS_SHUGART           2
+#define BUS_APPLE2            3
 
 
 /*
--- a/lib/usb/usb.proto
+++ b/lib/usb/usb.proto
@@ -3,7 +3,7 @@ syntax = "proto2";
 import "lib/common.proto";
 
 message GreaseWeazleProto {
-	enum BusType {
+	enum BusType { /* note that these must match CMD_SET_BUS codes */
 		BUSTYPE_INVALID = 0;
 		IBMPC = 1;
 		SHUGART = 2;


jepler:
The drive type numbers are officially allocated in greaseweazle (thanks @keirf!) and the software support is merged in the Adafruit_Floppy repo. So I think this is ready @davidgiven!

Flux writing is still not ready in Adafruit_Floppy, but when it's added `fluxengine write` will not need any further changes.

jepler:
Rebased to fix merge problems. It's mostly just a doc fix now so @davidgiven feel free to just close this up if it's superfluous.

davidgiven:
Yep, LGTM. Thanks!

keirf:
Looks dubious:
1. The actual change(s) in `usb.proto` seem to have got lost.
2. Only one new bus type remains in  `greaseweazle.h`. Fair enough, but I thought that was going to be the quarter-step variant, since the multiple stepping required to do whole or half steps can then be handled in software. But you have kept ID 3, which is the half-stepping variant (and that hasn't changed in Adafruit_Floppy afaics https://github.com/adafruit/Adafruit_Floppy/blob/main/examples/greaseweazle/greaseweazle.ino#L115).

So (1) this probably doesn't work at all; and (2) if it did, you wouldn't have the quarter-stepping variant.

Further, the whole exercise on the Greaseweazle side seems to be aimed only in reserving bus types, with the implementation and testing(!) on the FluxEngine side. This is actually fine by me, but should have been represented as such. The best way to handle that is to reserve IDs in the firmware's protocol header file, and to leave the Greaseweazle tools alone until/unless Apple2 support is properly introduced and tested.

In all this confusion: Which IDs do you actually want, and for what exact purpose?

jepler:
ID 3 for quarter stepping is good for us, I just merged the change in Adafruit_Floppy to work that way. The addition of ID 3 already happened in FE thanks to @davidgiven. I admit to not retesting on hardware today so it's possible something is wrong still.

keirf:
OK. But you should still test and fix the current merged fluxengine code.

jepler:
> Further, the whole exercise on the Greaseweazle side seems to be aimed only in reserving bus types, with the implementation and testing(!) on the FluxEngine side. This is actually fine by me, but should have been represented as such. The best way to handle that is to reserve IDs in the firmware's protocol header file, and to leave the Greaseweazle tools alone until/unless Apple2 support is properly introduced and tested.

I apologize about this, and wish I'd communicated better. Would it help if I investigated adding a2r flux writing support to Greaseweazle tools? Is there anything else I can do to improve things at this point, as far as greaseweazle goes?

davidgiven:
Does the Greaseweazle Apple II bus selection actually change anything? I don't have one, but I do have a Brother single-sided 3.5" drive with quad-stepping, and it needs to be provided with individual stepper motor phase lines. If instead the electrical interface is the same as a PC or Shugart drive, then I don't need the bus setting at all.

keirf:
I'm happy just to reserve an ID for now. It's not a problem - there are plenty!

The Apple II drive does have phase lines. This new bus type basically tells Adafruit_Floppy to use its somewhat different (than IBM/Shugart) predefined Apple II pinout.

jepler:
In Adafruit_Floppy it controls two things:
 * It uses a pinout with 4 individual motor phase lines instead of a pinout with step+direction
 * It uses a write waveform that is toggled rather than pulsed

jepler:
oh and the way that the write protect switch can be checked is nuts, you have to engage a particular winding of the stepper and then the write protect output is active :rofl:

jepler:
Tested with the tip of the master branch and the logical vs physical mapping seems to be working fine.

Typical commandline and output:
```
$ ./fluxengine read apple2 --usb.greaseweazle.bus_type=APPLE2 --usb.greaseweazle.port=/dev/ttyACM1 --flux_source.drive.high_density=0 -c 0-140x4
136.0: 200 ms in 61314 bytes
      40 raw records, 20 raw sectors; 4.03us clock (248kHz)
      sectors: 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
      logical track 34.0
      4096 bytes decoded
```

jepler:
Addendum: I wasn't testing what I thought I was, above.  Corrected:
```
fluxengine read apple2 apple2_drive  --usb.greaseweazle.port=/dev/ttyACM1 --drive.high_density=0 -c 0-34 --decoder.retries=0
...
136.0: 200 ms in 61256 bytes
      40 raw records, 20 raw sectors; 3.97us clock (252kHz)
      sectors: 34.0.0 34.0.1 34.0.2 34.0.3 34.0.4 34.0.5 34.0.6 34.0.7 34.0.8 34.0.9 34.0.10 34.0.11 34.0.12 34.0.13 34.0.14 34.0.15
      4096 bytes decoded
     Tracks -> 1         2         3         
H.SS 01234567890123456789012345678901234
0. 0 ...................................
0. 1 ...................................
0. 2 ...................................
0. 3 ...................................
0. 4 ...................................
0. 5 ...................................
0. 6 ...................................
0. 7 ...................................
0. 8 ...................................
0. 9 ...................................
0.10 ...................................
0.11 ...................................
0.12 ...................................
0.13 ...................................
0.14 ...................................
0.15 ...................................
Good sectors: 560/560 (100%)
Missing sectors: 0/560 (0%)
Bad sectors: 0/560 (0%)
      IMG: wrote 35 tracks, 1 sides, 140 kB total
```

However, now I'm unsure how to image all 140 quarter tracks into a raw flux format (#505) which is the most desirable for archiving copy protected media.

davidgiven:
The new group reader code will attempt to read the tracks in a group one-by-one, merging them together, until it gets a complete read. Then it'll skip to the next group as there's no point reading the rest of the group (it already has all the sectors). With an Apple II drive, there are four tracks in a group. You should be able to see it skipping tracks in the full log.

I'll add a flag to force it to read all tracks in the group --- it's very easy to do.

Issue #504, opened 2022-03-23
Can't "write" flux to `au:`?


jepler:
I noticed that I can't write to a kind of flux sink that isn't also a flux source. So for example, this fails but I think it should be possible for it to succeed:
```
$ ./fluxengine write ibm360_525 -i empty-360k.img -d au:noise/
Error: unrecognised flux source 'au:noise/'
```

davidgiven:
Hm. Yes, that's because of the setup for verification --- it's trying to configure the sink as a source as well so that the data can be read back. That ought to be a relatively simple fix...

davidgiven:
This should work now.

jepler:
thanks!

PR #505, opened 2022-03-23
Add apple-centric "a2r" raw flux format

### LINKED ISSUE #86: Apple: Support Applesauce A2R and WOZ
The Applesauce project (https://applesaucefdc.com) is a robust system for imaging Apple and Macintosh disks using the original drives. It uses a new format called WOZ for processed Apple II images. It also uses a format called A2R for flux images.

Both formats are specified here:
https://applesaucefdc.com/a2r/
https://applesaucefdc.com/woz/

Additionally, there is a Slack discussion board where the project (and working with Apple II and Macintosh disks and disk images in general) is discussed, which may be joined here: http://apple2.gs:3000

It would be great to support these formats, as they are quickly becoming the go-to for Apple II.

Example images in multiple formats may be found on the Internet Archive:
https://archive.org/details/flux_capacity

jepler:
With a commandline like the following:
```
$ ./fluxengine read apple2 --copy-flux-to apple.a2r apple2 appledos --flux_source.rescale=0.833
```
a raw flux file in the "A2R" format is written to `apple.a2r`. If its content is supported, it can be converted to the emulation-centric ".woz" file by [`passport.py convert apple.a2r`](https://github.com/a2-4am/passport.py), or even have passport.py remove its copy protection to create a dsk.

It may also be possible to use the .a2r file in the Applesauce software too, though I don't have a mac so I can't say for certain.

This actually works for an "Apple DOS 3.3" floppy, Adafruit's GW-compatible firmware, and a TEAC (pc-style) drive.

The dubious changes are related to https://github.com/davidgiven/fluxengine/issues/504 -- It isn't possible to `flux write -d SINK` unless SINK is also a source. I can drop them and work on #504 separately, but being able to `write` from an on-filesystem image while testing my new flux sink was very handy. (in retrospect I could have done a rawwrite from scp to a2r!)

Closes #86.


lib/fluxsink/a2rfluxsink.cc
```diff
@@ -0,0 +1,187 @@
...
+	void writeInfo() {
+		A2RInfoData info{
+		    A2R_INFO_CHUNK_VERSION,
+		    {}, // to be filled
+		    singlesided() ? A2R_DISK_525 : A2R_DISK_35,
+		    1,
+		    1,
+		};
+		char buf[33];
+		snprintf(buf, sizeof(buf), "Fluxengine %d%32.32s", FLUXENGINE_VERSION, "");
```
davidgiven, line 84: fmt::format() here instead of printf, please!

lib/fluxsink/a2rfluxsink.cc
```diff
@@ -0,0 +1,187 @@
...
+		std::stringstream ss;
+		for(auto &i : _metadata) {
+			ss << i.first << '\t' << i.second << '\n';
+		}
+		writeChunkAndData(A2R_CHUNK_META, ss.str().data(), ss.str().size());
+	}
+
+	void writeFlux(int cylinder, int head, const Fluxmap& fluxmap) override
+	{
+		// Note: this assumes that 'cylinder' numbers are in the PC high density 5.25 drive numbering, NOT apple numbering
```
davidgiven, line 99: This is due to change, but it's turning out to be more complicated than I thought --- I'll fix this myself after submission.

src/fe-write.cc
```diff
@@ -68,7 +68,8 @@ int mainWrite(int argc, const char* argv[])
 	if (argc == 1)
 		showProfiles("write", formats);
 	config.mutable_flux_sink()->mutable_drive();
-	config.mutable_flux_source()->mutable_drive();
+	if (verify)
```
davidgiven, line 71: Yep, this looks like a reasonable fix. It'd be nice to automatically invoke `--no-verify` if the sink isn't hardware-related but that can wait.

lib/fluxsink/a2rfluxsink.cc
```diff
@@ -0,0 +1,187 @@
...
+	{
+
+		std::cout << fmt::format("A2R: writing A2R {} file containing {} tracks\n",
+			singlesided() ? "single sided" : "double sided",
+			config.cylinders().end() - config.cylinders().start() + 1
+		);
+
+		char image_date[100];
+		time_t now{std::time(nullptr)};
+		strftime(image_date, sizeof(image_date), "%FT%TZ", gmtime(&now));
```
davidgiven, line 43: It _should_ be possible to do this with fmt::format --- look for strftime in https://github.com/fmtlib/fmt --- but I haven't tried it myself.

davidgiven:
Mostly looks good to me --- I've left some comments. Thanks!

lib/fluxsink/a2rfluxsink.cc
jepler: OK. is there a 'right' way to get the fluxengine software version? I think FLUXENGINE_VERSION is the device/protocol version? Or it can be left off.

lib/fluxsink/a2rfluxsink.cc
```diff
@@ -0,0 +1,187 @@
...
+class A2RFluxSink : public FluxSink
+{
+public:
+	A2RFluxSink(const A2RFluxSinkProto& lconfig):
+		_config(lconfig),
+		_bytes{},
+		_writer{_bytes.writer()}
+	{
+
+		std::cout << fmt::format("A2R: writing A2R {} file containing {} tracks\n",
```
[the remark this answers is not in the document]
jepler: oh yes will fix all these.

lib/fluxsink/a2rfluxsink.cc
davidgiven: Oops, missed that. FluxEngine doesn't really have a version, I'm afraid. FLUXENGINE_VERSION is, as you say, the protocol version and isn't related to the client.

src/fe-write.cc
jepler: is this related to / can it be solved by `has_drive()`?  I couldn't actually find where `has_drive` was implemented, it seemed to somehow come from the protobuf stuff.

davidgiven: Hmm. I hadn't thought of that. You should be able to test for a hardware sink with `config.flux_sink().has_drive()`, yes, once `parseFlags` is called.

jepler:
A friend tried applesauce on these files but got "failed to load! File contains an unknown kind of stream chunk.". I'll try to borrow a Mac for testing soon

davidgiven:
OK --- let me know how it goes... also, I should warn you that I've just done two gigantic refactors and you'll probably get a merge conflict when you update!

jepler:
This is now updated against the latest developments!

jepler:
@davidgiven can you take another look at this?  Together with https://github.com/adafruit/a2woz/ it now gives a good toolchain for archiving copy protected Apple II floppies and using them in an emulator such as mame.

lib/fluxsink/a2rfluxsink.cc
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+private:
+	void writeData(const auto data, size_t size) {
+		_writer += Bytes(reinterpret_cast<const uint8_t*>(data), size);
+	}
+	void writeChunkAndData(uint32_t chunk_id, const auto data, size_t size) {
+		_writer.write_le32(chunk_id);
+		_writer.write_le32(size);
+		writeData(data, size);
```
davidgiven, line 70: You can't write structures containing non-single-byte values like this. (The order in disk will vary depending on whether you're on a LE/BE system.)

davidgiven:
I did spot a couple of potential problems, but it mostly LGTM.

jepler:
Updated to address your review comments, thanks!

lib/fluxsink/a2rfluxsink.cc
tdaede: FWIW the D88 image reader also makes this assumption and will need to be adjusted if this is no longer the case.

jepler:
Hi! I'll be out for ~6 weeks but please let me know what needs to happen on this PR so it can get merged; I'll tend to it when I return, if testing on hardware is needed again.

davidgiven:
I'm sorry, this got dropped on the floor! Everything looks good and I'm merging. Thanks very much for doing this!

jepler:
Thank you!

PR #510, opened 2022-04-23
Allow an option to force the reader to read all tracks in a group.

davidgiven:
This is useful for archiving.

See #492.


jepler:
I'll try to give this a test but it looks like what I was looking for!  thank you

PR #517, opened 2022-05-30
Greaseweazle: Fix RPM measurement in some circumstances.

tdaede:
Accumulate number of ticks returned in FLUXOP_SPACE so that they
count towards the total track time.


davidgiven:
Thanks!

PR #522, opened 2022-05-31
Add D88 image writer.

tdaede:
Add D88 image writer.

This is currently a very minimal implementation and does not
support deleted sectors, CRC errors, etc.


davidgiven:
LGTM --- merged. Thanks!

PR #527, opened 2022-06-08
Switch build system to meson.

davidgiven:


meson.build
```diff
@@ -0,0 +1,195 @@
...
+protoc_generator = generator(
+  protoc,
+  output: ['@BASENAME@.pb.cc', '@BASENAME@.pb.h'],
+  arguments: ['--proto_path=@CURRENT_SOURCE_DIR@', '--cpp_out=@BUILD_DIR@', '@INPUT@'],
+)
+
+subdir('dep/agg')
+subdir('dep/fmt')
+subdir('dep/libusbp')
+subdir('dep/stb')
```
eli-schwartz, line 22: In general, you probably want to use https://mesonbuild.com/Wrap-dependency-system-manual.html here.

You can move dependencies to subprojects/ and write isolated meson.build files for them which export a sandboxed dependency interface and serve as a convention for including external dependencies.

Also for fmt specifically, you can just import it via `meson wrap install fmt`, switch to looking it up via `dependency()`, and seamlessly support debundled copies of the fmt library which may be available as part of the system on Linux.

eli-schwartz (changes requested):

PR #529, opened 2022-06-10
Imd reader

wybren1971:
Finally got around to adept the IMD reader to read an IMD image en set the protoconfiguration and be able to write it to physical media

### DIFF
--- a/lib/imagereader/imdimagereader.cc
+++ b/lib/imagereader/imdimagereader.cc
@@ -49,7 +49,7 @@ static unsigned getModulationandSpeed(uint8_t flags, bool *fm)
                 break;
 
             default:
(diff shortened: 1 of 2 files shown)


wybren1971:
please have a look at the code as i am not that familiar with proto. but it works fine with my tests

davidgiven:
All looks good to me, but just to check: did you mean to update the hplif770 format at the same time?

wybren1971:
Yes, because I found some problems with it. And this works better. 

I used this imdreader to write back an imd image to disk for my logic analyzer.

Thanks Wybren

davidgiven:
Excellent --- just checking. Thank you very much for doing this!

PR #541, opened 2022-06-20
add Imd writer option

wybren1971:
Changes in this pull request:
Add the ability to write imd images to file.
Update the documentation regarding the imd image possibilities.
make the HPLif format read and write
set the ibm format as default format in the gui version

### DIFF
--- a/lib/imagewriter/imdimagewriter.cc
+++ b/lib/imagewriter/imdimagewriter.cc
@@ -135,23 +135,19 @@ public:
 	void writeImage(const Image& image)
 	{
  		const Geometry& geometry = image.getGeometry();
-		unsigned numCylinders;
 		unsigned numHeads;
 		unsigned numSectors;
 		unsigned numBytes;
 		std::ofstream outputFile(_config.filename(), std::ios::out | std::ios::binary);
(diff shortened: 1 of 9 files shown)


lib/imagewriter/imdimagewriter.cc
```diff
@@ -0,0 +1,430 @@
...
+					numSectorsinTrack = 0;
+				nanoseconds_t RATE = 0;
+				if (sector->clock > 0)
+				{
+					RATE = 1000000.0 / sector->clock;
+				} else
+				{
+					switch (dataRate)
+					{
+					case ImdOutputProto::RATE_HD : RATE = 1000;
```
davidgiven, line 225: This probably want `break` statements!

davidgiven:
Thank you very much! LGTM, but I did spot one issue and left a comment...

wybren1971:
i added the breaks to the switch statement. spotted also one in the reader which i updated also

lib/imagewriter/imdimagewriter.cc
wybren1971: updates the code with break statements...

PR #549, opened 2022-08-10
Add n88basic sector gaps to d88 reader for PC-88 DD disks.

tdaede:
The values used for 1.2MB HD disks are too narrow for the
PC-88's disk controller.


davidgiven:
Merged --- thanks!

PR #550, opened 2022-08-10
D88: Fix sector ID conditions when writing.

tdaede:
This was failing to write the last sector in the image.


davidgiven:
Merged --- thanks!

PR #552, opened 2022-08-10
allow overlapping sectors

tdaede:
- Corrected spelling error
- Added real time and not the compiletime of fluxengine
- Allow new sectors to start in the data region of existing sectors.


davidgiven:
Sadly, this won't work --- seeking requires resetting the fluxmap reader state and causes data loss. This doesn't matter for architectures like the IBM where each record is distinct from the next but it caused a lot of problems on architectures like the Amiga where it's all one continuous data stream, so I had to rework it not to do that.

I think the only way to make this work is to try and detect a sync marker in the bitstream while reading the data and terminate the current record. That would simulate what a real FDC does, but FluxEngine's not really designed for that and it would be hard to do. In general I find that copy protected disks just aren't worth the effort...

tdaede:
Oh, I see. I guess I haven't done Amiga disks much but I would have expected them to appear as "one big sector" to this layer.

The actual FDC really doesn't look in the data stream either - for the ReadTrack command it's only reading record headers and skipping the body of the data.

Perhaps an easier fix is just for the Decoder to tell us whether its Records are individually seekable or not.

Note that tests actually pass here, which means we're probably missing some Amiga ones?

I am primarily working on copy protected disks now, if you consider those out of scope of the project it'd be good to know.

tdaede:
I implemented the "simple" fix. I disabled seeking for Amiga, though I could do it the other way around and only enable it for IBM.

PR #553, opened 2022-08-11
WIP: Add solved/reconstructed fluxmap support.

tdaede:
This proof of concept adds an optional encoder and fluxsink that
runs after the image is written.

This currently exposes the config of the solvedEncoder, though I
think I ultimately don't want it to be exposed as part of the
config proto.

Another way to do this would be by extending the decoders and/or
directly reading their records. That is actually closer to what I
want in many ways, however those records would need to be
deduplicated from multiple reads, etc.

The ultimate goal is to exactly replicate the timing of the
original sectors, but with a stable clock, sectors reread if a
bad read happens, and garbage between sectors removed. This
requires extending the sector struct to have more information
than it currently holds.


tdaede:
Moved solved encoder creation into arch/.

PR #554, opened 2022-08-18
Fix building on Mac OS X 10.13

abcbarryn:


abcbarryn:
These should allow the building of a Mac OS binary on OS X 10.13 (or newer) and build a Mac OS application package. I don't think it will break windows builds, other than removing tests from being built by default, but please test it as I do not currently have a Window build system. I also have a prebuilt Mac dmg file and command line installer in releases on my github page at: https://github.com/abcbarryn/fluxengine/releases/tag/FluxEngine-Mac-client-version-581

davidgiven:
As mentioned elsewhere (can't remember where, I hope you got it, github's code reviewing tool is weird) those `private`s are there intentionally as without them flags get propagated to various dependencies incorrectly, which can cause weird inconsistent builds. That makefile is supposed to be run with homebrew's `gmake`.

`makeapp.sh` looks really useful, though, and I've been dreading trying to find out how that stuff works. What's in the icon file --- AFAIK there is no official FluxEngine icon yet?

davidgiven:
I have actually added a properly OSX package, based on your stuff here --- thanks! The new GUI even has an icon!

So I'm closing this. Please feel free to reopen if there's anything else which needs doing...

PR #555, opened 2022-08-18
Remove private

abcbarryn:
private causes the build to fail on Mac OS X


davidgiven:
Sorry, but that's needed to prevent the value of CFLAGS from being propagated to the dependencies of the rule, which causes all kinds of horrible things from happening. Use gmake rather than make (it is in the documentation...).

PR #556, opened 2022-08-18
D88: Use the data length field to seek past any extra data.

tdaede:
This will effectively ignore the field length if it is shorter
than the sector size (or 0).


davidgiven:
I'll take your word for this one, as I know nothing about D88! I'll note the merge conflicts, though...

tdaede:
rebased

This matches the behavior of the M88 emulator and xdisk3.

davidgiven:
Thanks --- merged.

PR #588, opened 2022-09-13
One more attempt to fix the 1581 format...

davidgiven:


hpingel:
I did an incremental build and the writing of a known-good d81 image file to real disk fails on verify. I guess the check for swap_sides was kicked out of encoders.cc yesterday but didn't materialize somewhere else so it is not changing anything when put into the textpb profile.

davidgiven:
Damn, you're right, that doesn't work.

There is exactly one format which uses the side swapping feature, I don't have any known-good 1581 disks, I don't have a 1581, the encoder/decoder architecture isn't really designed to do this, and I tend to work on this in the evenings when I'm tired. Can you tell? I'll need to check in on this tomorrow, sorry.

Re moving collectSectors to the base class: this is because every architecture was implementing this itself, except the functionality is now replicated by the new layout stuff so a generic version will (mostly) work.

Issue #598, opened 2022-09-19
issues with --doc and IMD image writer


Yserra:
Hi David. I just compiled the latest source codes from Github (on september 19th, 2022), and I saw so far two issues. 
My configuration is Mac M1 macOS v12.6

The --doc option doesn't write anymore a documentation. It waits for ever without ending. Only pressing CTRL-C can return to terminal.

```
$ ./fluxengine read --doc
^C
$ ./fluxengine read amiga --doc
^C
$ ./fluxengine read amiga -o essai.adf --doc
^C

```
Reading from an IBM DD 3.5" disk, 720 KB formatted (one I just created with "fluxengine write ibm720", by the way), to an .IMD image doesn't seem to work as expected. 
I tried: 

```
$ ./fluxengine read ibm -o essai720.imd
Measuring rotational speed...       Using GreaseWeazle GW000F002A4E30501420343135 on /dev/cu.usbmodem1201
200.3ms (299.5rpm)
Error: LAYOUT: no sectors in sector definition!
```

then:
```

$ ./fluxengine read ibm720 -o essai720.imd
Measuring rotational speed...       Using GreaseWeazle GW000F002A4E30501420343135 on /dev/cu.usbmodem1201
200.3ms (299.5rpm)
 0.0: 240 ms in 66362 bytes
      39 raw records, 10 raw sectors; 2.00us clock (500kHz)
      sectors: 0.0.1 0.0.2 0.0.3 0.0.4 0.0.5 0.0.6 0.0.7 0.0.8 0.0.9
      4608 bytes decoded
...
```
(all the disk is read, but, at the end, the file is 0 KB...)
```
...
Good sectors: 1440/1440 (100%)
Missing sectors: 0/1440 (0%)
Bad sectors: 0/1440 (0%)
      IMD: guessing data rate as RATE_DD
      IMD: guessing recording mode as RECMODE_MFM
Error: IMD: Can't write IMD files with this speed 2008, and modulation false. Try another format.

```

During building of the sources, I got this error:

```
lib/imagewriter/imdimagewriter.cc:228:14: warning: enumeration value 'ImdOutputProto_DataRate_RATE_GUESS' not handled in switch [-Wswitch]
                                        switch (dataRate)
                                                ^~~~~~~~
lib/imagewriter/imdimagewriter.cc:228:14: note: add missing switch cases
                                        switch (dataRate)
                                                ^
```

davidgiven:
[...]

PR #600, opened 2022-09-21
bug fix in data rates

wybren1971:
#598 the flags in this function where not adjusted what resulted in the mentioned error. Hope this fixes it.


davidgiven:
Thanks!

PR #604, opened 2022-10-02
Force density setting for D88/NFD.

### LINKED ISSUE #560: D88 reader does not correctly set low/high density automatically
It is currently always high density by default.

tdaede:
This fixes #560 (partially).

I do not understand how the protobuf parsing stuff works enough
to make it detect optional fields.

### DIFF
--- a/lib/imagereader/d88imagereader.cc
+++ b/lib/imagereader/d88imagereader.cc
@@ -65,19 +65,15 @@ public:
         int clockRate = 500;
         if (mediaFlag == 0x20)
         {
-            Logger() << "D88: high density mode";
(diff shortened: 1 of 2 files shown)


davidgiven:
Reasonable, yeah. The original thought was that the user-supplied settings shouldn't override the autodetected ones but I don't think that'll work because the formats themselves will be likely to supply these settings.

tdaede:
I would at some point like to turn the high_density setting into a 3 way setting - high, low, and auto. For auto it would come from the format, and when reading it would switch densities by reading the first track and looking at the recovered clock rate.

PR #605, opened 2022-10-02
D88/NFD: Add physical layout before adding sector.

### LINKED ISSUE #603: Writing D88 with `ibm` format is broken
```
Error: LAYOUT: no sectors in sector definition!
```

tdaede:
Partially fixes #603, it is still busted if the number of
tracks is specified larger than what is in the image, which
used to work.


tdaede:
Also it's still broken with this change as it only will write a maximum of 1 sector per track.

davidgiven:
Bah. Yes --- Image.put now requires a valid layout...

PR #607, opened 2022-10-02
Remove the layout cache and rename layout -> trackInfo.

### LINKED ISSUE #603: Writing D88 with `ibm` format is broken

tdaede:
The layout cache is busted and seems not worth the effort.

Fixes #603.

### DIFF
--- a/lib/layout.cc
+++ b/lib/layout.cc
@@ -4,9 +4,6 @@
 #include "lib/environment.h"
 #include <fmt/format.h>
 
-static Local<std::map<std::pair<int, int>, std::shared_ptr<TrackInfo>>>
-    layoutCache;
-
 static unsigned getTrackStep()
 {
     unsigned track_step =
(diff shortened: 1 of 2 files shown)


davidgiven:
This one I'm not sure about. This did actually seem to significantly improve performance (although I'll admit I never actually bothered to benchmark it). What problems were you seeing?

tdaede:
Without this, if I write a d88 file with the `ibm` format, it only writes one sector onto every track.

tdaede:
Also, I'm not seeing anything that clears the cache if you write 2 successive disks (e.g. via the GUI).

davidgiven:
I'm sorry, I dropped this one.

The cache is actually cleared by magic via calling `Environment::reset()`, which will call `clear()` on all globals wrapped in the `Local<>` construct. However, if you're actually seeing problems then it clearly doesn't work, and as it's just a performance booster, let's disable it.

Issue #625, opened 2022-12-01
Mac version of FluxEngine (GUI) won't start – libraries?


elosha:
Hey David + team,

I wanted to try FluxEngine with my GreaseWeazle and opted for the GUI version. I took the latest prebuilt version. Noticed it comes with several of the required libraries in its .app bundle, as well as a script that handles fallback to these libraries in case they are not installed on the system. This is great! However I noticed …

1. The default search path seems to be for Homebrew package management (`/usr/local/opt/…/lib/…`), they don't work on systems maintained by MacPorts (`/opt/local/lib/…`)

So I would love to fallback to use of those bundled libs. But …

2. The DYLD_FALLBACK_FRAMEWORK_PATH variable from .app bundle is ignored, an error that the libraries can't be found at `/usr/local/opt/…` still occurs.

Help :(

**My system:**
MacPro5,1 with Intel Xeon X5650
Mac OS "High Sierra" 10.13.6 (latest)
MacPorts package manager + definitions 2.8.0 (latest)

davidgiven:
So, being honest here: I'm not an OSX programmer. I stole all that stuff from someone else without really understanding what I was doing. I have found a few things I could try...

Since you've obviously gone inside the application package: could you try putting all the libraries in the Contents/Resources directory into a Contents/Resources/lib directory and seeing if that does anything?

elosha:
That provided the same result. But I noticed the bundle contains just libraries which it tries to locate via `DYLD_FALLBACK_FRAMEWORK_PATH`, while this variable would fit better: `DYLD_FALLBACK_LIBRARY_PATH`.

Once I define this, the problem is gone … and then the real trouble begins:

> dyld: Library not loaded: /usr/local/opt/libpng/lib/libpng16.16.dylib
>   Referenced from: /Applications/FluxEngine.app/Contents/Resources/libwx_osx_cocoau_core-3.2.dylib
>   Reason: image not found
>

elosha:
Since I've patched the fallback paths now, there's just one thing left: The diverging filenames.

This script looks up which files have been linked and need to be copied to the .app bundle:
[fluxengine/src/gui/build.mk](https://github.com/davidgiven/fluxengine/blob/master/src/gui/build.mk):58

[...]

PR #626, opened 2022-12-01
Library fallback path fixed & MacPorts compatible

elosha:
Use correct fallback variable for libs, not for frameworks. Also look for libs in MacPorts' default lib path /opt/local/lib, not only HomeBrew's


davidgiven:
Even better than good bug reports --- bug fixes!

I think I would rather include my own libpng (does OSX really not provide one by default?) but this will do for now. Thank you!

davidgiven:
Ref: #625

PR #663, opened 2023-04-23
Micropolis: disk rotate period is 200 ms

ejona86:
The disks are expected to contain 100,000 bitcells, so clock_period_us and rotational_period_ms need to align.


davidgiven:
Merged --- thanks.

PR #664, opened 2023-05-01
Adding IBM PC 8-sector formats

bdwheele:
I've come across some early IBM PC disks that use 8-sectors per track rather than the more common 9.  This branch adds the format so I can manipulate them.


davidgiven:
LGTM, thanks! Also, weird...

PR #690, opened 2023-06-25
Fix incorrect product name in installer.

Deledrius:
This fix corrects the product name shown on the initial installer page, as well as the name of the Desktop shortcut created so that it matches the name provided in the Uninstall script.


davidgiven:
Whoops --- thanks!

PR #702, opened 2023-07-29
Micropolis: Fix encoder and decoder to support encodedecodetest

ejona86:
These changes should not impact reading/writing from real disks. This includes a bug fix to Fluxmap where it might miss a trailing interval when adding bits, as mentioned in #333.

With the Fluxmap bug fixed, the encoder now includes index pulses in its output. The decoder was relaxed to allow reading precisely one track.

We don't actually add an encodedecodetest for micropolis, though, because the SCP encoder is unhappy with so many revolutions.

### DIFF
--- a/arch/micropolis/decoder.cc
+++ b/arch/micropolis/decoder.cc
@@ -85,9 +85,10 @@ public:
         /* Discard a possible partial sector at the end of the track.
          * This partial sector could be mistaken for a conflicted sector, if
          * whatever data read happens to match the checksum of 0, which is
-         * rare, but has been observed on some disks.
+         * rare, but has been observed on some disks. There's 570uS of slack in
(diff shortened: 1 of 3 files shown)


davidgiven:
I think SCP only supports up to five revolutions, and doesn't support hard sectors at all? I'm a bit out of touch with what's happened to the format recently.

Also, there needs to be a way to allow testing with just `.flux` files. Anyway, merged --- thanks.

ejona86:
I've got SCP hacked locally to support more. It's "change the magic number 5" and "insert index pulse between tracks." I have been keeping it to myself because it isn't clear it should be changed. Hard sectors are just "really strange track rotations." At least that's what greaseweazle spit out in #322 and what fluxengine does without encouragement.

PR #703, opened 2023-07-29
Micropolis: Add missing HCS order for VGI

ejona86:


davidgiven:
Merged, thanks (also, if you have a known good flux image and VGI file for a blank Micropolis disk, it can be added to https://github.com/davidgiven/fluxengine-testdata to check for further regressions).

PR #706, opened 2023-08-05
Makefile: Eagerly run pkg-config for protobuf

ejona86:
Protobuf added a dependency on absl and now pkg-config is incredibly slow. `pkg-config --libs protobuf` and `--cflags` each take around 1.5 seconds on my laptop. Running pkg-config only once reduces a 100% incremental build for 'make all' from 90 seconds to 3.2 seconds.

Unfortunately we will pay the 3 seconds every time we execute make, even for something that doesn't need protobuf.


davidgiven:
Sigh. Thanks.

ejona86:
FYI: pkgconf 2.0, released two weeks ago, fixes the performance issue (.07s). So at least there's hope.

PR #709, opened 2023-08-14
Make the non-gui executable build on FreeBSD

bsdphk:
Here is a minimal patch to make the non-gui executable compile on FreeBSD.

I created a "dummy.c" file for libusbp with sufficient entrypoints for compilation to succeed but all of them failing.


davidgiven:
Merged --- thanks!

PR #717, opened 2023-09-26
Fix const-correct/discards qualifiers error.

### LINKED ISSUE #696: fluxengine client failing to build on linux / unsure how to get gui on linux?
I dwnloaded both the current state of the github, as well as the source code on the latest release, both encounter the same issue after extraction and running make in the proper directory.

```
[make command issued]
[PROTOC and CXX lines]
In file included from src/fe-analyselayout.cc:6:
lib/sector.h: In member function ‘auto fmt::v6::formatter<Sector::Status>::format(Sector::Status, fmt::v6::format_context&) const’:
lib/sector.h:99:48: error: passing ‘const fmt::v6::formatter<Sector::Status>’ as ‘this’ argument discards qualifiers [-fpermissive]
   99 |             Sector::statusToString(status), ctx);
      |                                                ^
In file included from lib/globals.h:18,
                 from src/fe-analyselayout.cc:2:
/usr/include/fmt/format.h:2942:8: note:   in call to ‘decltype (ctx.out()) fmt::v6::formatter<T, Char, typename std::enable_if<(fmt::v6::internal::type_constant<T, Char>::value != fmt::v6::internal::custom_type), void>::type>::format(const T&, FormatContext&) [with FormatContext = fmt::v6::basic_format_context<std::back_insert_iterator<fmt::v6::internal::buffer<char> >, char>; T = fmt::v6::basic_string_view<char>; Char = char; typename std::enable_if<(fmt::v6::internal::type_constant<T, Char>::value != fmt::v6::internal::custom_type), void>::type = void; decltype (ctx.out()) = std::back_insert_iterator<fmt::v6::internal::buffer<char> >]’
 2942 |   auto format(const T& val, FormatContext& ctx) -> decltype(ctx.out()) {
      |        ^~~~~~
make: *** [Makefile:271: .obj/src/fe-analyselayout.o] Error 1
```
I am able to compile version 389 (after installing ninja, which is not on the list of dependencies) [Though now i'm lost as that seems to only be a command line program? I'm here for a GUI, and im only seeing a cmd executable after running make]

kristomu:
This should fix #696 and let gcc compile FluxEngine.

### DIFF
--- a/lib/sector.h
+++ b/lib/sector.h
@@ -93,7 +93,7 @@ struct Sector : public LogicalLocation
 template <>
 struct fmt::formatter<Sector::Status> : formatter<string_view>
 {
-    auto format(Sector::Status status, format_context& ctx) const
+    auto format(Sector::Status status, format_context& ctx)
     {
         return formatter<string_view>::format(
             Sector::statusToString(status), ctx);


davidgiven:
Merged, thanks (although, oddly, I use GCC for development and it builds fine).

Issue #722, opened 2023-10-30
make install gone?


VorpalBlade:
It appears `make install` has been removed, which breaks the AUR package build.

Also `make binaries` is gone, and `make tests` apparently does nothing?

Note that AUR PKGBUILDS have the following general structure:

```bash
build() {
	cd "$srcdir/${pkgname%-git}"
	# Make all depends on tests, but we want to run them in check() instead.
	make binaries
}

check() {
	cd "$srcdir/${pkgname%-git}"
	make tests
}

package() {
	cd "$srcdir/${pkgname%-git}"
	make install PREFIX=/usr DESTDIR="$pkgdir/"
	install -Dm 644 COPYING.md "$pkgdir/usr/share/licenses/$pkgname/COPYING.md"
	# Installing docs is not yet supported in a reasonable way by upstream.
}
```

Since you have your own home grown build system it is kind of hard to figure out how to do this now after the recent changes.

davidgiven:
Oh, whoops. I forgot you were using that. Will fix.

VorpalBlade:
Hm, doesn't actually seem to work yet:

```
==> Starting package()...
INSTALL
'.obj/src+fluxengine/src+fluxengine' -> '/build/fluxengine-git/pkg/fluxengine-git//usr/bin/fluxengine'
install: cannot create regular file '/build/fluxengine-git/pkg/fluxengine-git//usr/bin/fluxengine': No such file or directory
make: *** [Makefile:59: install-bin] Error 1
```

Not sure if it is on my side or your side, but too late for me to look at it today.

davidgiven:
Whoops. Forgot to add the `-D` to create the directory. Fixed (or will be as soon as the build completes).

Issue #721, opened 2023-10-23
Fluxengine GUI unable to open any any image for writing


misterblack1:
Hi, when I try to select an image to write with FluxEngine GUI I always get "Error: cannot open input file"

I tested by creating a new image off a real 1.44mb MSDOS floppy using Fluxengine GUI, saving it as ibm.img. Then when trying to write it back to the disk, selecting the image I just created gives the error. (Or any other image.) 

I am above to use FluxEngine GUI to browse the 'ibm.img' file and see the contents, so I know this isn't a permission problem with reading the file.

Windows 10 64 + GreaseWeazle + using the latest automated build

![Screenshot 2023-10-23 135848](https://github.com/davidgiven/fluxengine/assets/16782710/514462f3-ef2f-4b99-a92f-98a7bff400eb)

davidgiven:
Yes, that's a bug in the GUI which I'm looking at. The command line version works fine, though.

misterblack1:
Ah ok thanks!

Indeed I was able to write with the command line -- although I'm also running into the write precompensation issue as well.  (I saw another open issue. I can't write IBM images to 3.5" disks with the greaseweazle, although 360k disks do seem to work OK, like Apple II floppies.)

Issue #701, opened 2023-07-29
Build system doesn't seem to respect '-j' for parallel builds


VorpalBlade:
It seems that the build system mostly runs serially even though I'm passing `-j30`. It also seems to invoke `pkg-config --cflags protobuf` way too many times (again serially). It really only should need to invoke that once.

I would attempt to debug this, but since you use a home grown Makefile based solution (rather than autoconf, cmake, meson or something similar) I'd rather not.

I'm on commit ab9d6cf5edfc40921c454c10789a5b905241f068.

davidgiven:
All I can say is that it works for me. I'm on a 24-core system and do parallel builds as a matter of course.

Regarding build systems... I'd love to use an established one, but every one I've established so far falls down in some way. autoconf/automake is almost impossible to make valid parallel makefiles with, meson isn't scriptable, bazel can't make mingw32 binaries (believe it or not, it's hard coded to use mingw64 exclusively) and in general is very hostile to open source builds, etc. cmake is a possibility but I'm having trouble getting past the awful scripting language; plus, it doesn't do combined release/debug builds, which I'd really like to do.

VorpalBlade:
I might look into this (probably won't be for over a week at least). Do you have any documentation of how the build system is set up? I.e. what the overall design of it is.

Also, I *think* parallel builds used to work for me. It seems like a new thing that they broke. I use Arch Linux (and packaged fluxengine for [AUR](https://aur.archlinux.org/packages/fluxengine-git)). It was when rebuilding fluxengine against newer system dependencies that I noticed the slow build and started investigating. Possibly due to being on a rolling release distro I noticed some issue sooner than you would have done.

davidgiven:
It's a fairly simple non-recursive makefile using implicit patterns to actually build things based on variable names, automake-like. The only complex bit is the use of macros to add dependencies. src/build.mk shows how to build a binary, lib/build.mk shows a library. The main rules which actually do the work are in the top-level Makefile. It relies heavily on rule-specific variables.

There are bugs; sometimes it won't rebuild things when it should, and I need to sort those out. As I said, I'd love an alternative, but so far haven't seen anything. I wish the meson author would add proper scripting; it's so nearly exactly what I want...

eli-schwartz:
> I wish the meson author would add proper scripting; it's so nearly exactly what I want...

By proper scripting, do you mean functions?

davidgiven:
[...]

PR #723, opened 2023-10-30
Miscellaneous bugfixes.

davidgiven:
Put back the standard makefile targets.
Fixes: #722 

The GUI can open flux and image files for writing.
Fixes: #721 

Make the build system saner.
Fixes: #701 

Also, strip out the broken tpi stuff in favour of something much simpler (and more working).


misterblack1:
Hi David, unfortunately the GUI disk writing still isn't working for me with this latest build. (fb9f7fe from 45 minutes ago)  Same error on picking any disk image to write.  

I tested with 1440k IBM format images and a Mac 800k disk image. 

image

Windows 10.0.19045.3570 + Greaseweazle

davidgiven:
Huh. I've just tried it and it works here. Can you describe your exact process?

misterblack1:
I made a short video:
https://youtu.be/b4I3EudYpTU

The error is the same with IBM disks -- hadn't tried others but I assume it's the same with all. As you mentioned before, the command line work fine.

chironb:
I, too, I still seeing this issue with the GUI program. I'm using the 2-day old release, so I know it's the latest version. I'm getting the same "cannot open file" error. I've also tried installing it as administrator and I've run the program as administrator. Same error. Just like the video. Doesn't seem to matter what formats I select or what file I choose. 

I also tried the FluxEngine command-line version. I got this output: 

.\fluxengine.exe write C:\Users\micro\Desktop\System3.2_Journal.Mac400.img

[libprotobuf ERROR ../../src/google/protobuf/text_format.cc:335] Error parsing text-format ConfigProto: 1:3: Message type "ConfigProto" has no field named "LK".

Error: couldn't load external config proto

I hope that helps!

chironb:
In addition to my earlier comment, I have a little bit of possibly useful observations. In my comment I listed the problem with the command-line version and saw it was related to the google/protobuf/ stuff. But in the GUI version the error seems to be generated here: 
            wxMessageBox(e.message, "Error", wxOK | wxICON_ERROR);
            if (_state == STATE_WRITING_WORKING)
                SetState(STATE_WRITING_FAILED);
That seems to imply that auto filename = wxFileSelector works but then 
            if (filename.empty())
                return;
returns, which gets us to the final error message. 

So it seems like there is some problem with the library used to either get the file from Windows 10 (in my case) or a problem with something that maybe is in google/protobuf but I'm not sure. 

BUT!!!!

When I choose a file called Untitled.png it gives me the error: 
unrecognized image filename
C:\Users\micro\Desktop\Untitled.png

So... I took an floppy disk image file and renamed it to Untitled.png just to see what would happen. It gives the same error as the actual file that was an actual .png file. 

This is also true with files like .exe or .jpg in that whenever I try to open them, it clearly gets a full path and filename but then looks at the extension and decided it's not an actual image filename. Because I would guess that if it were reading inside the file it would figure out that the wrongly named disk image file would have worked. 

Clearly it's getting as far as analyzing the filename to determine if it's a valid image, but then failing when it tries to start reading the actual file. 

I'm using Greaseweazle with FluxEngine and reading works just fine. But this seems more like an issue with FluxEngine and local file access and does not seem to even get as far as accessing Greaseweazle, since the drive doesn't make any noise or get accesses in any way and the error message seems to be presented instantly. 

I hope this helps even if it's maybe too much information!

chironb:
One more detail: I can format a disk using the format button in the FluxEngine GUI.

PR #749, opened 2024-04-02
Set DTR after calling SetCommState

p-j-b:
After calling SetCommState in setBaudRate DTR is no longer set. 
The amendment in this request clears and sets DTR again, like in the constructor.

### DIFF
--- a/lib/usb/serial.cc
+++ b/lib/usb/serial.cc
@@ -97,6 +97,12 @@ public:
             .Parity = NOPARITY,
             .StopBits = ONESTOPBIT};
         SetCommState(_handle, &dcb);
+
+        if (!EscapeCommFunction(_handle, CLRDTR))
+            error("Couldn't clear DTR: {}", get_last_error_string());
+        Sleep(200);
+        if (!EscapeCommFunction(_handle, SETDTR))
+            error("Couldn't set DTR: {}", get_last_error_string());
     }
 
 private:


p-j-b:
When DTR is not set the Windows Fluxengine application is not sent any more data from Adafruit Floppy  Greaseweazle (which uses the TinyUSB library).
In this use case it will appear like the fluxengine application has _"frozen"_ after  the Greaseweazle port informaton has been displayed, as fluxengine is waiting for ReadFile to receive data that will not be sent. For example:

```
fluxengine.exe read atarist --800 -o c:\temp\st.img --usb.greaseweazle.port=COM14
      OPTION: 800kB 3.5" 80-track 10-sector DSDD
      Measuring rotational speed...
      Using Greaseweazle on serial port COM14
```

This is only an issue for the Windows application as calling cfsetspeed function does not clear DTR.

I have confirmed that amendment works when using the fluxengine application with a Raspberry Pico running the Adafruit Floppy Greaseweazle example sketch.  However, I don't have access to actual Fluxengine or Greaseweazle hardware so I haven't been able to test this amendment against any other hardware.

davidgiven:
Oops; just spotted this... thanks very much. This could actually explain an issue I've been seeing on Linux with a real Greaseweazle. It'll need porting to OSX and Linux but I'll take care of that.

davidgiven:
Oh, wait, it _is_ ported! Curses. That means the issue is something else...

PR #789, opened 2024-11-07
Update from upstream

FoamyGuy:
@ladyada 

I attempted to resolve the conflicts, I'm unsure if I've done so correctly without breaking the DSK type stuff that you added previously.

A few specific bits I don't know about

1) In imagewriter.proto I changed
```
optional DskOutputProto dsk = 8
```
to 
```
optional DskOutputProto dsk = 10
```
because upstream introduced a different type assigned at `8`.

2) In imagewriter.cc `ImageWriter::updateConfigForFilename` was removed from the upstream, but ours had been modified to add the .dsk stuff. I brought it up from ours into this version, but I don't know whether we truly need it still or if there would need to be change elsewhere to hand dsk now.


FoamyGuy:
wrong repo, sorry

PR #799, opened 2024-12-21
fix broken screenshot link

foobarbecue:
Path was relative so it was looking for doc/doc/screenshot-details.png instead of doc/screenshot-details.png, causing a 404.


davidgiven:
I can't believe it took me so long to get to this... in fact, the fix was very slightly different; done elsewhere. Thanks for the report.

PR #800, opened 2025-01-29
Expose standalone .app target for osx (macOS)

jedahan:
This makes it easier to create a homebrew formula


build/ab.py
```diff
@@ -507,7 +507,7 @@ def export(self, name=None, items: TargetsMap = {}, deps: Targets = []):
             cwd=self.cwd,
             ins=[srcs[0]],
             outs=[destf],
-            commands=["$(CP) %s %s" % (srcs[0], destf)],
+            commands=["$(CP) -r %s %s" % (srcs[0], destf)],
```
jedahan, line 510: since a `.app` file is actually a directory this is needed.

davidgiven:
Sorry this took so long to get to --- that's a good idea, but unfortunately it can't be done like that (directories can't be build targets). But I have changed the build system to expose a FluxEngine.app.zip which contains the FluxEngine.app.

Issue #804, opened 2025-03-17
Build failure with protobuf 3.30


VorpalBlade:
I get the following build error after Arch Linux updated to protobuf 3.30:

```
lib/config/proto.cc: In function ‘std::string getProtoFieldValue(ProtoField&)’:
lib/config/proto.cc:291:35: error: could not convert ‘enumvalue->google::protobuf::EnumValueDescriptor::name()’ from ‘google::protobuf::internal::DescriptorStringView’ {aka ‘std::basic_string_view<char>’} to ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’}
  291 |             return enumvalue->name();
      |                    ~~~~~~~~~~~~~~~^~
      |                                   |
      |                                   google::protobuf::internal::DescriptorStringView {aka std::basic_string_view<char>}
lib/config/proto.cc: In instantiation of ‘findAllProtoFields(google::protobuf::Message*)::<lambda(auto:34*, const auto:35&)> [with auto:34 = const google::protobuf::Descriptor; auto:35 = std::__cxx11::basic_string<char>]’:
/usr/include/c++/14.2.1/type_traits:2655:26:   required by substitution of ‘template<class _Fn, class ... _Args> static std::__result_of_success<decltype (declval<_Fn>()((declval<_Args>)()...)), std::__invoke_other> std::__result_of_other_impl::_S_test(int) [with _Fn = findAllProtoFields(google::protobuf::Message*)::<lambda(auto:34*, const auto:35&)>&; _Args = {const google::protobuf::Descriptor*, const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&}]’
 2655 |       std::declval<_Fn>()(std::declval<_Args>()...)
      |       ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14.2.1/type_traits:2666:60:   required from ‘struct std::__result_of_impl<false, false, findAllProtoFields(google::protobuf::Message*)::<lambda(auto:34*, const auto:35&)>&, const google::protobuf::Descriptor*, const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&>’
 2666 |       using type = decltype(_S_test<_Functor, _ArgTypes...>(0));
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
[...]

PR #811, opened 2025-06-21
Fix Linux compile issues related to protobuf

TheBombSquad:
Couldn't compile this on Linux with Protobuf 3.31, seems to be related to issue #804 where string_view could not be implicitly converted to std::string. This fixes the build by making the conversions explicit.


davidgiven:
Thank you for this --- sorry it took me so long to get to it. I've applied the patch separately due to conflicting with other changes. Unfortunately, I don't have Protobuf 31 yet (all the platforms I build for are still on 19 or so) so I can't test it directly. Let me know if anything seems to break?

Issue #795, opened 2024-11-29
CP/M filesystem fails for disks with non standard track order


boamaod:
I want to browse my [DSDD "outout" disks](https://github.com/infoaed/juku3000/blob/master/docs/kettad.md#millega-ja-mispidi-neid-siis-lugeda) from [JUKU E5104](https://en.wikipedia.org/wiki/Juku_E5101) and it seems there is no configuration option to support either that or "outback" (in [Libdsk parlance](https://www.mankier.com/5/libdskrc#Parameters)). Currently I make it work by recalculating the track positions at `vfs.cc` [just before obtaining them](https://github.com/davidgiven/fluxengine/blob/master/lib/vfs/vfs.cc#L299-L301):

```
if (sides>1)
{
    track *=2;
    if(side==0)
    {
        if (track>=tracks) { side++; track-=tracks; }
    }
    else
    {
        if (track<tracks) { side--; track++; }
        else { track-=tracks-1; }
    }
}
```

If one would like to convert it to a configuration option, there might be a better way to do it though. Is it a feature worth adding and how to fit it in? Number of 800kB DSDD "outout" disks for testing can be found [here](https://elektroonikamuuseum.ee/failid/juku/tarkvara/).

davidgiven:
There are some image layout options in lib/layout.proto, but not this one (currently you can specify CHS and HCS).

To confirm I'm thinking about the same thing you are: does this affect the position of tracks _on the disk_, or just the order of the blocks in .dsk files?

boamaod:
I tried CHS/HCS, but did not fully understand how they relate to my case. They didn't appear to directly help me with my "outout" floppies in any combination.

About the ontological question, the tracks are _on JUKU and some other disks_ in that order and need to remain so for the system and the emulators, but in order to make sense of the filesystem, they have to be processed in the correct order by third party tools. Alternating tracks between sides as the default way of reading a floppy disk normally makes tracks appear in natural order also on disk image and this image can be directly read as the basis for CP/M filesystem. For "outout" and "outback" you have to either read the tracks in different order, convert the disk or reorder the tracks on the fly to meaningfully process them.

[...]

PR #825, opened 2025-08-27
Correct Juku E5104 documentation

boamaod:
There was a correction discussed at #795, but never implemented. Also find a non-copyrighted flux image of a Juku floppy in [the last message](https://github.com/davidgiven/fluxengine/issues/795#issuecomment-3219126881) of the discussion. And also fixed the links to refer to non GitHub versions of referred documents on public Internet (https://j3k.infoaed.ee/).


davidgiven:
Excellent, thanks!

Issue #829, opened 2025-11-20
Cannot compile with Debian 13 (trixie)


Turboschlumpf:
make
Environment changed --- regenerating
AB
loading build.py
loading build/c.py
loading build/utils.py
loading build/toolchain.py
loading build/protobuf.py
loading build/pkg.py
loading config.py
fluxengine-testdata not found; skipping corpus tests
loading dep/fmt/build.py
loading src/build.py
loading dep/adflib/build.py
loading dep/agg/build.py
loading dep/alphanum/build.py
loading dep/fatfs/build.py
loading dep/hfsutils/build.py
loading dep/libusbp/build.py
loading dep/stb/build.py
loading lib/config/build.py
loading lib/core/build.py
loading lib/external/build.py
loading lib/decoders/build.py
loading arch/build.py
loading lib/fluxsink/build.py
loading lib/encoders/build.py
loading lib/fluxsource/build.py
loading lib/imagereader/build.py
loading lib/imagewriter/build.py
loading lib/usb/build.py
loading lib/vfs/build.py
loading lib/data/build.py
loading dep/lexy/build.py
loading lib/algorithms/build.py
loading src/formats/build.py
loading scripts/build.py
loading tests/build.py
loading src/gui2/build.py
Error materialising Target('+all'): <function export at 0x7f4bd12f6c00>
Arguments: {'self': Target('+all'), 'name': '+all'}
Traceback (most recent call last):
  File "/home/dk/fluxengine/fluxengine-dev/build/ab.py", line 761, in <module>
    main()
    ~~~~^^
  File "/home/dk/fluxengine/fluxengine-dev/build/ab.py", line 751, in main
    t.materialise()
    ~~~~~~~~~~~~~^^
  File "/home/dk/fluxengine/fluxengine-dev/build/ab.py", line 359, in materialise
    raise e
  File "/home/dk/fluxengine/fluxengine-dev/build/ab.py", line 323, in materialise
    v = t.convert(v, self)
  File "/home/dk/fluxengine/fluxengine-dev/build/ab.py", line 463, in convert
    output = {k: target.targetof(v) for k, v in value.items()}
                 ~~~~~~~~~~~~~~~^^^
  File "/home/dk/fluxengine/fluxengine-dev/build/ab.py", line 378, in targetof
    return targetof(value, self.cwd)
  File "/home/dk/fluxengine/fluxengine-dev/build/ab.py", line 436, in targetof
    loadbuildfile(path)
[...]

PR #830, opened 2025-11-30
Update ab.

davidgiven:
Builds should now work on 3.10 and above.

Fixes: #829


davidgiven:
Fixes: #829 (hopefully)

PR #849, opened 2026-02-01
Add build documentation for RHEL and Fedora

qbarnes:
I added the necessary package list for building fluxengine for RHEL and Fedora.  I'm not sure if there's anywhere else in the docs that also need to be updated to reflect this change.

I also found a redundant package in the Ubuntu package list.  I deleted it in a separate commit.

Feel free to do anything at all you want to the commits.


davidgiven:
Belatedly merging. Thanks!

PR #855, opened 2026-05-05
smaky6fs: fix size bug, 0xFF deleted entries, decode directory metadata

Sch-LikA:
Three bugs in `lib/vfs/smaky6fs.cc` found by cross-referencing the original Smaky 6 hardware documentation and independent reverse-engineering of disk images (including Winchester hard-disk images not previously analysed).

## Bug 1 — Wrong file size when last sector is completely full

When `last_bytes == 0` the last sector holds a full 256 bytes. The old formula:
```c
length = (endSector - startSector - 1) * 256 + lastSectorLength;
```
gives a result **256 bytes short** in that case. Fixed to:
```c
length = lastSectorLength
    ? (endSector - startSector - 1) * 256 + lastSectorLength
    : (endSector - startSector) * 256;
```

## Bug 2 — Deleted entries with first byte `0xFF` were not skipped

The original check `if (dbuf[0])` only skips empty slots (first byte = `0x00`). On real Smaky 6 disks and all known Winchester hard-disk images, deleted/system entries use `0xFF` as a sentinel. This caused those entries to be parsed as live files with garbage field values.

Fixed to: `if (dbuf[0] && dbuf[0] != 0xff)`

## Bug 3 — Directory bytes `[0x0e–0x17]` were treated as unknown

These fields are fully decodable (documented in the original Smaky 6 hardware manual and confirmed from disk image analysis):

| Offset | Field | Notes |
|--------|-------|-------|
| `0x0e–0x0f` | `flags` | uint16 LE, purpose TBD |
| `0x10–0x11` | `last_bytes` | bytes used in last sector (0 = full) |
| `0x12–0x13` | `load_addr` | high byte then low byte; reliable for type `SM` |
| `0x14–0x15` | `entry_addr` | same encoding as load_addr |
| `0x16` | `month` | BCD (e.g. `0x07` = July); `0x00` or `0xFF` = no date |
| `0x17` | `year` | BCD (e.g. `0x82` = 1982); year ≥ 78 → 19xx, else 20xx |

The raw `smaky6.metadata_bytes` attribute is replaced by the properly decoded `smaky6.flags`, `smaky6.load_addr`, `smaky6.entry_addr`, and `smaky6.date` attributes.

### DIFF
--- a/lib/vfs/smaky6fs.cc
+++ b/lib/vfs/smaky6fs.cc
@@ -5,9 +5,19 @@
 
 /* A directory entry looks like:
  *
- * 00-09: ten byte filename FFFFFFFF.EE
- * 0a-0b: word: start sector
- * 0c-17: unknown
+ * 00-07: name, space-padded ASCII (high bit may be set as attribute flag → mask with 0x7f)
+ * 08-09: 2-char type code (SY, SM, ST, SR, LS, FH, BS, IM, KS …)
+ * 0a-0b: uint16 LE start_sector  – first data sector (0-based absolute)
+ * 0c-0d: uint16 LE end_sector    – first sector after the file (exclusive)
+ * 0e-0f: uint16 LE flags         – purpose unknown
+ * 10-11: uint16 LE last_bytes    – bytes used in last sector;
+ *                                   0 means last sector is completely full
+ *         exact_size = (size_sectors-1)*256 + last_bytes  if last_bytes > 0
+ *                      size_sectors * 256                  if last_bytes == 0
+ * 12-13: load address  – high byte then low byte (reliable for type SM)
+ * 14-15: entry point   – same encoding
+ * 16:    month BCD (0x07 = July);  0x00 or 0xFF = no date
+ * 17:    year  BCD (0x82 = 1982);  0x00 or 0xFF = no date
  */
 
 class Smaky6Filesystem : public Filesystem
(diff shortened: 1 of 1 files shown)


lib/vfs/smaky6fs.cc
```diff
@@ -132,23 +135,20 @@ class Smaky6Filesystem : public Filesystem
...
-                if (dbuf[0] && dbuf[0] != 0xff)
-                {
-                    auto de = std::make_shared<SmakyDirent>(dbuf);
-                    dirents.push_back(de);
-                }
-            }
+        /* Sub-directory inside a DR container:
+         * The first 3 sectors of the DR entry hold the sub-directory.
+         * Sector numbers stored there are relative to drStartSector. */
+        Directory(Smaky6Filesystem* fs, unsigned drStartSector)
```
davidgiven: You could annotate this as =0 and not need the other constructor.

lib/vfs/smaky6fs.cc
```diff
@@ -200,37 +219,85 @@ class Smaky6Filesystem : public Filesystem
...
+        if (path.empty())
+        {
+            /* List the root directory. */
+            Directory dir(this);
+            for (auto& de : dir.dirents)
+                result.push_back(de);
+            return result;
+        }
+
+        if (path.size() == 1)
```
davidgiven: Does the filesystem not support nested subdirectories?

lib/vfs/smaky6fs.cc
```diff
@@ -200,37 +219,85 @@ class Smaky6Filesystem : public Filesystem
...
 
-        return dir.findFile(path[0]);
+        throw BadPathException(path);
     }
 
     Bytes getFile(const Path& path) override
     {
-        if (path.size() != 1)
-            throw BadPathException(path);
+        if (path.size() == 1)
```
davidgiven: The directory lookup code here is repeated multiple times. It'd probably be better to have a single method which looks up a path and call it from all these places.

davidgiven: (Actually, getDirent should already do this?)

Sch-LikA:
Thanks for the review! I've pushed a new commit ([a4a56dd](https://github.com/Sch-LikA/fluxengine/commit/a4a56ddd)) addressing all three points:

**Constructor redundancy** — merged the two `Directory` constructors into one with `unsigned drStartSector = 0`. Sector 0 is the root, so the default covers both cases cleanly without a separate overload.

**Repeated path-traversal code** — added two private helpers:
- `directoryAt(path)` — returns the `Directory` for the path's parent level (root for depth 0, DR sub-directory for depth 1, throws `BadPathException` beyond that).
- `resolveDirent(path)` — calls `directoryAt` on the parent path then `findFile` on the last component.

`list()`, `getDirent()`, and `getFile()` each now delegate to these helpers — all the traversal logic lives in one place, and `getDirent` effectively drives the shared resolution as you suggested.

**Nested sub-directories** — the Smaky 6 filesystem is structurally flat-within-one-level: a `.DR` entry is a container whose first three sectors hold a plain 32-entry directory; the spec does not allow a `.DR` inside another `.DR`. So depth > 1 is correctly rejected with `BadPathException`. The new `directoryAt` helper documents this constraint explicitly in its comment.

davidgiven:
LGTM. Thanks!

PR #857, opened 2026-05-10
Makefile: Fix auto-detection of Windows environment

Sonic-Amiga:
Plain "make" now works correctly in MSY2/Mingw64 shell for Windows build. No more need to explicitly provide BUILDTYPE=windows.


davidgiven:
Merging. Thanks!

PR #859, opened 2026-05-11
Introduce TR-DOS disk format support

Sonic-Amiga:
TR-DOS is a built-in ROM-based operating system for BetaDisk interface for the ZX-Spectrum by Technology Research Ltd. In late 80s it was brought into (back-then) USSR and cloned, having become a de-facto standard floppy disk interface for ZX-spectrum clones.
Official documentation says both 40- and 80- single- and double- sided drives are supported, however in practice i never met any other formats than 80-track DS. The controller is off-the-shelf 1793 chip, so encoding scheme it a standard IBM-compatible MFM. Sector size is 256 bytes, 16 sectors per track.
Official extension for plain TR-DOS disk images is .trd, so i added it to the list of recognized extensions by IMG writer.

Tested with real hardware by successfully reading a disk, written from .trd image using Fluxengine with Greaseweasle.


Sonic-Amiga:
@davidgiven Hello! I have brought up my real hardware and successfully tested disks, written using this patch. Works like a charm! Changed write support to Unicorn, ready for merge.

davidgiven:
Sorry for the long delay in getting back to you --- I've been rewriting the entire client; I've ported your changes to the new version. See #877. Thanks!

PR #863, opened 2026-06-23
Switch from unique_ptr/shared_ptr to garbage collection.

davidgiven:


davidgiven:
Couldn't make it work, now obsolete.

